PR #32, opened 2016-06-02
Add timeout option

misterMuyiwa:
This commit adds an optional timeout param into the easypost library.
This gives users the ability to specify what timeout value they would
like to use when making requests to the Easypost API.

If a timeout is not passed in, a default timeout of 120000ms is used.


victoryftw:
Awesome! Thanks for this, its a nice feature!

PR #38, opened 2016-12-09
v3.0.0

ajacksified:
* Test coverage!
* Up to date with currently documented APIs!
* Similar API to v2 (and non-node APIs) for minimal changes!
* Modern es6 and our internal .eslint file!
* Promises; yield and await friendly!
* A repl!

Things left to do:

* [x] Hook up with CI
* [x] Draft easypost.com/docs/api docs
* [ ] Integration test suite - run tests against prod api with test API key
* [x] Endpoints
  * [x] API keys
  * [x] Carrier accounts
  * [x] Users


package.json
```diff
@@ -1,22 +1,54 @@
...
   "author": "Sawyer Bateman <support@easypost.com>",
   "homepage": "https://easypost.com",
   "repository": {
     "type": "git",
     "url": "git://github.com/easypost/easypost-node.git"
   },
-  "main": "lib/main.js",
+  "main": "index.js",
   "engines": {
     "node": ">= v0.4.0"
```
ajacksified: With the babel shim, should work anywhere; nobody should be on node 0.4 anymore, though. I'd be happy with bumping to at least 0.10. I did play around with producing a node 6+ version, which runs less compiler rules and produces a smaller and faster file, but the effects were pretty minimal. I may revisit and publish two files later on; one could include `require('node-easypost')` or `require('node-easypost/node-6`) depending.

package.json
```diff
@@ -1,22 +1,54 @@
...
-    "test": "node test/main.js"
+    "test": "mocha --require ./test/common.js --compilers js:babel-register --recursive ./test",
+    "test:coverage": "cross-env NODE_ENV=test nyc mocha --recursive ./test",
+    "build": "webpack --config webpack.config.js",
+    "lint": "eslint src",
+    "watch": "webpack --watch --config webpack.config.js"
+  },
+  "dependencies": {
+    "proptypes": "^0.14.3",
+    "superagent": "^3.1.0"
```
ajacksified: :+1:

repl.js
```diff
@@ -0,0 +1,95 @@
...
+    'the repl with an API_KEY environment variable.'
+  ].join(' '));
+}
+
+const local = repl.start('$> ');
+
+function help () {
+  const help = [
+    'To try out the API, use the available instance of `api` to make requests.',
+    'For example, try writing: `api.Address.all();',
```
ajacksified: Whoops, thanks.

src/resources/base.js
```diff
@@ -0,0 +1,171 @@
...
+    // suppressVAlidation is used when creating objects from API responses-
+    // the API returns keys that we don't later on use for creating or editing.
+    // We want access to read these, but we don't want to throw validation
+    // errors; they're valid, but read-only. Note to self: maybe add a readonly
+    // type that uses getters and setters?
+    constructor(data = {}) {
+      this.mapProps(data);
+    }
+
+    validateProperties(throwOnFailure = true) {
```
ajacksified: It's reading its properties as an instance, so it shouldn't be static (I suppose it _could_ be static and have the proptypes and properties passed in, but I don't see a benefit.)

test/resources/pickup.js
```diff
@@ -0,0 +1,169 @@
+/* eslint import/no-named-default: 0 */
```
ajacksified: Yeah; I don't think it's a particularly useful rule. I'll make it global.

package.json
```diff
@@ -1,22 +1,54 @@
 {
   "name": "node-easypost",
   "description": "EasyPost Node Client Library",
-  "version": "2.1.1",
+  "version": "3.0.0",
   "author": "Sawyer Bateman <support@easypost.com>",
```
ajacksified: Maybe I'll change it to "Easypost Engineering" or something, and point at the support email?

src/resources/base.js
```diff
@@ -0,0 +1,171 @@
...
+          if (!args[i]) {
+            throw new Error(`Missing parameter: ${p}`);
+          }
+        });
+      }
+    }
+
+    async rpc(path, body) {
+      const slashPath = path ? `/${path}` : '';
+      const url = `${this.constructor._name.toLowerCase()}/${this.id}${slashPath}`;
```
eschutho: cool, good to know.

src/resources/shipment.js
```diff
@@ -0,0 +1,158 @@
+import T from 'proptypes';
```
k-funk: $> api.Shipment.all().then(console.log)
Promise { <pending> }
$> TypeError: this.unwrapAll(...).map is not a function

Tracker has a `unwrapAll` fn, this one doesn't. I assume this is missing that function? Insurances, and some other objects should probably have that fn as well.

PR #39, opened 2016-12-19
hook up /reports endpoint

k-funk:


src/resources/payment_log_report.js
```diff
@@ -0,0 +1,8 @@
+import reportBase from './report_base';
+
+export default api => (
+  class PaymentLogReport extends reportBase(api) {
+    static _name = 'PaymentLogReport';
+    static url = 'reports/payment_log';
```
k-funk: Can the `reportBase` class have `'reports/'` that these classes bring in?

ajacksified: It could, if you refactored a little like this:


```javascript
export default api => {
  const Base = reportBase(api);

  return class PaymentLogReport extends Base {
    static _name = 'PaymentLogReport';
    static url = `${Base.url}/payment_log`;
    // ...
  }
}
```

k-funk: ah. I was doing `this.url` instead of `Base.url`. thanks!

ajacksified:
👍

PR #44, opened 2017-01-19
/scan_forms endpoint (api.ScanForm)

k-funk:
@ajacksified object creation doesn't match the objects that are returned.

curl:
`curl -X POST https://api.easypost.com/api/v2/scan_forms -u <api_key>: -d '{"shipments":[{"id":"shp_59515f02dfb84fbe972c94f69bbb2aec"}]}' -H 'Content-Type: application/json'`

Few questions:
Do I need `jsonIdKeys`? What is that field for?

Should I be overriding `toJSON` so that `new api.ScanForm({shipments: [{id: "shp_59515f02dfb84fbe972c94f69bbb2aec"}]})` works?

### DIFF
--- a/test/resources/scan_form.js
+++ b/test/resources/scan_form.js
@@ -1,0 +1,50 @@
+import apiStub from '../helpers/apiStub';
+import NotImplementedError from '../../src/errors/notImplemented';
+
+import scanForm from '../../src/resources/scan_form';
+
+describe('ScanForm Resource', () => {
+  it('exists', () => {
+    expect(scanForm).to.not.be.undefined;
+    expect(scanForm).to.be.a('function');
+  });
+
+  it('unwraps array responses', () => {
+    const ScanForm = scanForm(apiStub());
+    const data = [new ScanForm()];
+    expect(ScanForm.unwrapAll({ scan_forms: data })).to.deep.equal(data);
+  });
+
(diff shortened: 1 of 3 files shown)


ajacksified:
`jsonIdKeys` takes embedded objects, such as customsInfo on a shipment, and sends only the id instead of the entire object, if an id exists. It's only necessary if you have embedded objects.

Yes, you should override tojson.

test/resources/scan_form.js
```diff
@@ -0,0 +1,45 @@
...
+
+    it('returns the shipments if available', () => {
+      const shipments = { shipments: [{ id: 'shp_123' }] };
+      const sf = new ScanForm(shipments);
+      expect(sf.toJSON()).to.deep.equal(shipments);
+    });
+
+    it('calls the super if there\'s no shipments', () => {
+      const sf = new ScanForm();
+      expect(sf.toJSON()).to.deep.equal({});
```
k-funk: @ajacksified while this passes, it's not quite correct. it doesn't actually test that the super was called. Don't see any examples of that in the current tests.

ajacksified: Good callout - this should spy on super instead of just checking the result.

ajacksified:
Code looks good - do you want to add a `super` test?

PR #47, opened 2017-01-21
bring coverage back to 100%

k-funk:


test/easypost.js
```diff
@@ -40,6 +40,14 @@ describe('Base API object', () => {
 
       expect(API.buildHeaders(merge)).to.deep.equal(mergedHeaders);
     });
+
+    it('removes headers when in a browser', () => {
+      const filteredHeaders = { ...DEFAULT_HEADERS };
+      delete filteredHeaders['User-Agent'];
+      delete filteredHeaders['Accept-Encoding'];
+
```
k-funk: need to mock this so `typeof window !== 'undefined'`. thoughts on how to do this?

PR #48, opened 2017-01-24
fix object creation. requires object key in singular form, not plural

k-funk:


src/resources/parcel.js
```diff
@@ -5,6 +5,7 @@ export default api => (
   class Parcel extends base(api) {
     static _name = 'Parcel';
     static url = 'parcels';
+    static key = 'parcel';
```
ajacksified: parcels will need a custom tojson - they shouldn't be wrapped in posts. (reports are the other endpoint that works this way.)

PR #51, opened 2017-02-09
Wrap parcel json

ajacksified:
@k-funk  Do you know why the wrapping key was removed here? https://github.com/EasyPost/easypost-node/commit/cb71fe15dd020a1b1b3b7a884b9eb7c7c6383864#diff-83a975aa8cf501970485ca625ed9df32R22

This fixes an open ticket about parcel creation.


k-funk:
I think parcel creation used to work. Think we should write some more tests for resource-specific regression testing?

PR #62, opened 2017-03-22
parameterized Report type instead of BaseReport/ShipmentReport...

k-funk:


src/resources/report.js
```diff
@@ -20,12 +19,33 @@ export default api => (
       updated_at: T.object,
     }
 
+    constructor(data = {}) {
+      super(data);
+      if (data.type) this._url = this.constructor.constructUrl(data.type);
```
ajacksified: single-line ifs need braces (should be a linter rule)

k-funk: linter isn't catching it. looked for that eslint setting in their docs, but didn't find it.

PR #69, opened 2017-05-08
Allow passing in service and carrier to automatically buy a shipment

ajacksified:
For #68 

Covers an undocumented usecase - providing service and carrier automatically buys the shipment. Also allows for passing in an array of services for rate loookups.


ajacksified:
No, this is for save.

PR #77, opened 2017-05-30
Fix api key child mapping

ajacksified:
This was improperly reducing, and created nested arrays.

### DIFF
--- a/src/resources/apiKey.js
+++ b/src/resources/apiKey.js
@@ -30,7 +30,7 @@ export default api => (
       let res = data.keys.map(k => ({ ...k, user_id: data.id }));
 
       if (data.children && data.children.length) {
-        res = res.concat(data.children.map(d => this.convertKeyMap(d)));
+        res = res.concat(data.children.reduce((arr, d) => arr.concat(this.convertKeyMap(d)), []));
       }
 
       return res;
--- a/test/resources/apiKey.js
+++ b/test/resources/apiKey.js
@@ -100,9 +100,6 @@ describe('ApiKey Resource', () => {
         }],
       };
 
-      console.log(data);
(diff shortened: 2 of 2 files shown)


src/resources/apiKey.js
```diff
@@ -30,7 +30,7 @@ export default api => (
       let res = data.keys.map(k => ({ ...k, user_id: data.id }));
 
       if (data.children && data.children.length) {
-        res = res.concat(data.children.map(d => this.convertKeyMap(d)));
+        res = res.concat(data.children.reduce((arr, d) => arr.concat(this.convertKeyMap(d)), []));
```
k-funk: this change seems equivalent...?

test/resources/apiKey.js
```diff
@@ -75,4 +71,43 @@ describe('ApiKey Resource', () => {
...
+
+          children: [{
+            id: 3,
+            keys: [{ c: '3' }],
+          }],
+        }],
+      };
+
+      console.log(data);
+      console.log(ApiKey.convertKeyMap(data));
```
k-funk: remove these?

src/resources/apiKey.js
ajacksified: The difference here: `data.children.map()` returns an array of whatever the function returns, one for each element in `data.children`, and this.convertKeyMap() returns an array. So, with children, you end up with an array of arrays, one for each child.

With `reduce`, we are concatenating each instance of convertKey map into a single array that we pass in - adding what's _inside_ of convertKeyMap to a top-level array, and returning that.

That said, I could save a bit of code and change this line to

`data.children.reduce((arr, d) => arr.concat(this.convertKeyMap(d)), res)`

Since what we really want to do is concat the result of the child convertKeyMap call with res anyways.

PR #84, opened 2017-07-05
Added Rating Endpoint

algreen11:
Can now call rating endpoint, example found below (important to call console.dir with depth of 4, as otherwise the rates just show up as [Array]), also not entirely sure why lines 48 and 49 were changed in package.json, as I didn't do that manually. 
```
const rt = new api.Rating({
  to_address: {
  name: 'Dr. Steve Brule',
    street1: '179 N Harbor Dr',
    city: 'Redondo Beach',
    state: 'CA',
    zip: '90277',
    country: 'US',
    phone: '4155559999',
    email: 'dr_steve_brule@gmail.com'
},
  from_address: {
  name: 'EasyPost',
  street1: '118 2nd Street',
  street2: '4th Floor',
  city: 'San Francisco',
  state: 'CA',
  zip: '94105',
  phone: '415-123-4567'
},
  parcels: [{
  length: 20.2,
  width: 10.9,
  height: 5,
  weight: 65.9
}],
  carrier_accounts:["ca_f5de031c879e4a4faebb7a7dcbf5dc7c", "ca_5496e6901d3240ffa2b1ba17674fb46f"]
});

rt.save().then((r) => {console.dir(r, {depth: 4})}).catch((error) => {console.log(error);});
```

### DIFF
--- a/src/easypost.js
+++ b/src/easypost.js
@@ -119,7 +119,7 @@ export default class API {
     return this.baseUrl + path;
   }
 
-  async request(path = '', method = METHODS.GET, params, headers = {}) {
+  async request(path = '', method = METHODS.GET, params = {}, headers = {}) {
     const {
       query,
       body,
@@ -133,8 +133,8 @@ export default class API {
 
     if (path === 'rating/v1/rates') {
       req.url = 'https://api.easypost.com/rating/v1/rates';
-      req['_header']['content-type'] = 'application/json';
-      req['header']['Content-Type'] = 'application/json';
+      req._header['content-type'] = 'application/json';
+      req.header['Content-Type'] = 'application/json';
     }
 
     if (body) { req.send(body); }
--- a/src/easypost.js
+++ b/src/easypost.js
@@ -130,9 +130,9 @@ export default class API {
                 .set('Content-Type', 'application/json')
                 .set(API.buildHeaders(headers))
                 .auth(`${this.key}:`);
-                
+
     if (path === 'rating/v1/rates') {
-      req.url =  'https://api.easypost.com/rating/v1/rates';
+      req.url = 'https://api.easypost.com/rating/v1/rates';
       req['_header']['content-type'] = 'application/json';
       req['header']['Content-Type'] = 'application/json';
     }
--- a/src/easypost.js
+++ b/src/easypost.js
@@ -26,7 +26,7 @@ import RequestError from './errors/request';
 
 export const MS_SECOND = 1000;
 export const DEFAULT_TIMEOUT = 120 * MS_SECOND;
-export const DEFAULT_BASE_URL = 'https://api.easypost.com/v2/';
+export const DEFAULT_BASE_URL = 'https://api.easypost.com/';
 
 export const UA_INFO = {
   client_version: pkg.version,
@@ -127,15 +127,10 @@ export default class API {
 
     const req = this.agent[method](this.buildPath(path))
                 .accept('json')
-                .set('Content-Type', 'application/json')
+                .type('json')
+                .set('content-type', 'application/json')
                 .set(API.buildHeaders(headers))
                 .auth(`${this.key}:`);
-
-    if (path === 'rating/v1/rates') {
-      req.url = 'https://api.easypost.com/rating/v1/rates';
-      req._header['content-type'] = 'application/json';
-      req.header['Content-Type'] = 'application/json';
-    }
 
     if (body) { req.send(body); }
 
--- a/src/easypost.js
+++ b/src/easypost.js
@@ -14,6 +14,7 @@ import Insurance from './resources/insurance';
 import Order from './resources/order';
 import Parcel from './resources/parcel';
 import Pickup from './resources/pickup';
+import Rating from './resources/rating';
 import Report from './resources/report';
 import ScanForm from './resources/scan_form';
 import Shipment from './resources/shipment';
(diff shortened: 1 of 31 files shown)


src/easypost.js
```diff
@@ -123,10 +126,16 @@ export default class API {
...
-                  .accept('json')
-                  .set('Content-Type', 'application/json')
-                  .set(API.buildHeaders(headers))
-                  .auth(`${this.key}:`);
+                .accept('json')
+                .set('Content-Type', 'application/json')
+                .set(API.buildHeaders(headers))
+                .auth(`${this.key}:`);
+
+    if (path === 'rating/v1/rates') {
```
ajacksified: Why does this need to be explicitly set? It should be hitting thus URL anyways, and the content-type is set to application/json above.

src/resources/base.js
```diff
@@ -153,20 +153,24 @@ export default api => (
...
       try {
         const data = this.constructor.wrapJSON(this.toJSON());
-
+        let output;
         let res;
 
         if (this.id) {
           res = await api.put(`${this._url || this.constructor._url}/${this.id}`, { body: data });
+          output = this;
+        } else if (this.constructor._name === 'Rating') {
```
ajacksified: Instead of changing code in resources/base to make changes specific to rating, the `rating` endpoint  should override the `save` method.

`body` is already set to `this.constructor.wrapJson(this.toJSON())`, so why do you need to set it to `this.toJSON` here again as a separate case?

Instead of setting output to `this.body`, it should be set to `this`, because `this` should have its attributes merged with the api response.

src/resources/rating.js
```diff
@@ -0,0 +1,32 @@
...
+    static propTypes = {
+      to_address: T.oneOfType([T.string, T.shape(Address.propTypes)]),
+      from_address: T.oneOfType([T.string, T.shape(Address.propTypes)]),
+      parcels: T.arrayOf(T.oneOfType([T.string, T.shape(Parcel.propTypes)])),
+      carrier_accounts: T.array,
+    }
+
+    static all() {
+      return this.notImplemented('all');
+    }
```
ajacksified: Needs newlines after each method's closing brace.

test/resources/rating.js
```diff
@@ -0,0 +1,33 @@
...
+    });
+  });
+
+  it('throws on retrieve', () => {
+    const Rating = rating(apiStub());
+    return Rating.retrieve().then(() => {}, (err) => {
+      expect(err).to.be.an.instanceOf(NotImplementedError);
+    });
+  });
+});
```
ajacksified: If you _do_ end up needing special logic to cover `save` for some reason, it needs tests here.

src/easypost.js
algreen11: `Request {
  domain: null,
  _events: { end: { [Function: bound onceWrapper] listener: [Function: bound _clearTimeout] } },
  _eventsCount: 1,
  _maxListeners: undefined,
  _agent: false,
  _formData: null,
  method: 'POST',
  url: 'https://api.easypost.com/v2/rating/v1/rates',
  _header: 
   { 'user-agent': 'EasyPost/v2 NodejsClient/3.0.12',
     accept: 'application/json',
     'content-type': 'application/x-www-form-urlencoded',
     'accept-encoding': 'gzip,deflate,sdch,br',
     'x-easypost-client-user-agent': '{"client_version":"3.0.12","lang":"nodejs","lang_version":"v8.1.2","publisher":"easypost","platform":"darwin"}',
     authorization: 'Basic OWY2elFXcVZrNlpQN2xFUGRSYjZyQTo=' },
  header: 
   { 'User-Agent': 'EasyPost/v2 NodejsClient/3.0.12',
     Accept: 'application/json',
     'Content-Type': 'application/x-www-form-urlencoded',
     'Accept-Encoding': 'gzip,deflate,sdch,br',
     'X-EasyPost-Client-User-Agent': '{"client_version":"3.0.12","lang":"nodejs","lang_version":"v8.1.2","publisher":"easypost","platform":"darwin"}',
     Authorization: 'Basic OWY2elFXcVZrNlpQN2xFUGRSYjZyQTo=' },
  writable: true,
  _redirects: 0,
  _maxRedirects: 5,
  cookies: '',
  qs: {},
  qsRaw: [],
  _redirectList: [],
  _streamRequest: false }`

algreen11: ^^ for some reason, the content type is set to urlencoded even so (the above code is when I comment out the content type change), also the base url is https://api.easypost.com/v2 , but the url required for rating is just https://api.easypost.com/rating/v1/rates

src/resources/base.js
algreen11: constructor.wrapJson makes the input "rating : {input}", but we just need "{input}"

if I set it to this.body, when printing out the rating call, it gives me the addresses, parcels and then ratings, but the way its supposed to return is just ratings without the address and parcel

src/easypost.js
ajacksified: I see what's going on - switch `set('Content-Type', 'application/json')` to `.type('json')` at line 130.

ajacksified: > also the base url is https://api.easypost.com/v2 , but the url required for rating is just https://api.easypost.com/rating/v1/rates

Let's change the base to `https://api.easypost.com` then, and update the endpoints to include the `v2` prefix. The base class shouldn't have anything specific to individual endpoints - all the code for each endpoint should be contained entirely within its own class.

algreen11: weird - I was only able to get it by setting this combo
```
.type('json')
.set('content-type', 'application/json')
```

Justintime50:
Standalone rating is no longer supported. Thank you for your efforts on this PR!

PR #87, opened 2017-11-01
Do not add an extra : superagent takes care of this for us

orloc:


ajacksified:
Interesting - looks like superagent@3.5.2 updated the auth function to look like this:

```javascript
Request.prototype.auth = function(user, pass, options){
  if (1 === arguments.length) pass = '';
  if (typeof pass === 'object' && pass !== null) { // pass is optional and can be replaced with options
    options = pass;
    pass = '';
  }
  if (!options) {
    options = {
      type: 'function' === typeof btoa ? 'basic' : 'auto',
    };
  }

  var encoder = function(string) {
    if ('function' === typeof btoa) {
      return btoa(string);
    }
    throw new Error('Cannot use basic auth, btoa is not a function');
  };

  return this._auth(user, pass, options, encoder);
};
```

3.5.2 isn't even tagged in their GitHub anymore, so they must consider it a broken release. I'll bump the packages and update to use the new auth signature. Thanks for noticing this!

orloc:
Sweet Thanks so much !

ajacksified:
Closed in favor of #88.

PR #102, opened 2018-07-13
Add proxy support

yonilerner:


ajacksified:
Hi @yonilerner,

I took your idea and implemented it a little more generically, so that clients can choose their own superagent wrappers. Let me know what you think here: https://github.com/EasyPost/easypost-node/pull/104

PR #115, opened 2018-12-20
Adding examples folder with basic request examples

LoganSimonsen:
I scrubbed all files of API keys and real addresses. I might want to add some more example files in the future, but for now I think this is a good list to start with. No changes or deletions from other folders or files.


k-funk:
Gonna close this in favor of pointing users to https://www.easypost.com/docs/api/node for sake of only having one place to maintain code samples.

PR #125, opened 2019-10-17
make retrieval id optional

cwidafor:


k-funk:
Thanks for the PR. We've decided to solve this User-specific problem a different way in the next day or two.

PR #127, opened 2019-12-11
Upgrade build packages

ajacksified:
Bumps packages to latest, specifically for NPM audit fixes.


package.json
```diff
@@ -35,40 +35,40 @@
...
-    "vows": "0.8.2",
+    "nyc": "^14.1.1",
+    "sinon": "^7.3.2",
+    "sinon-chai": "^3.3.0",
+    "vows": "^0.8.2",
     "webpack": "^4.39.3",
-    "webpack-cli": "^3.3.7",
-    "webpack-node-externals": "1.7.2"
+    "webpack-cli": "^3.3.10",
+    "webpack-node-externals": "^1.7.2"
```
k-funk: curious, why were only some deps pinned?

ajacksified: We used to pin everything, and a few things probably slipped through. There's no need for pinning in libraries, though, so I've removed all of the pins.

PR #132, opened 2020-03-11
Removed shipment return

Justintime50:
Shipments cannot be returned simply by calling the ID. Documentation has already been corrected but the client library needs this item removed as well. The documentation example (which has since been removed) looked like this:

```javascript
// Alternatively, return an instance you've already loaded.
api.Shipment.retrieve('shp_...').then(s => {
  s.return().then(() => console.log(s));
});
```

As this isn't supported and the documentation has been corrected, this should also be removed from the client library.

There was conversation around whether to just release this (breaking change), or wait for another fix and sneak it in together.

### DIFF
--- a/src/resources/shipment.js
+++ b/src/resources/shipment.js
@@ -127,16 +127,6 @@ export default api => (
       return this.rpc('refund');
     }
 
-    async return() {
-      this.verifyParameters({
-        this: ['id', 'to_address', 'from_address'],
-      });
-
-      const data = this.toJSON();
-      data.is_return = true;
-      return this.rpc('', data);
-    }
-
     lowestRate(carriers, services) {
       let rates = this.rates || [];
 
--- a/test/resources/shipment.js
+++ b/test/resources/shipment.js
@@ -277,65 +277,6 @@ describe('Shipment Resource', () => {
     });
   });
 
-  describe('return', () => {
-    let Shipment;
-    let si;
-    let stub;
-
-    beforeEach(() => {
-      stub = apiStub();
-      Shipment = shipment(stub);
-      si = new Shipment({ id: '1' });
-    });
-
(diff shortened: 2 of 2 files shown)


k-funk:
This should probably be a major version bump. We are considering whether or not there are other features/breaking changes in the near future to do at the same time.

Justintime50:
@k-funk sounds good, this can sit here ready until then.

Justintime50:
I'm realizing I've compiled the library in this PR, let me know if you'd like me to remove those and just have the code changes here and then compile/update version in a following PR.

att14:
If we don't want to checkin the `easypost.*` files we should add them to the `.gitignore`.

Was this ever working or was it broken from the beginning?

Justintime50:
@att14 we could, but we don't have any workflow for building new versions currently - especially if they were ignored. Is this correct?

I don't believe we've ever been able to return a shipment in this manner. The code on the commit before this was introduced did not have any mention of it (https://github.com/EasyPost/easypost-node/blob/eee118c329c07066c844f019d21480e65293f98c/lib/main.js#L386), commit `eaa88b72bd0ae02ec0e580bc63b14b755c9ebf26` introduced this method which to my knowledge never worked.

k-funk:
> If we don't want to checkin the `easypost.*` files we should add them to the `.gitignore`.
> 
> Was this ever working or was it broken from the beginning?

`easypost.*` files should be checked in. Modules published to npm should be transpiled (to vanilla js) so that consumers don't run into node version issues between what easypost-node maintains its src files in, and what they use. Other packages sometimes organize them into `/dist`.

Every time a version bumps, they should be updated. If you see version bumps in the commit history without those transpiled files being updated, those are mistakes.

att14:
If this never worked I'd say it doesn't need to be a major release. If it did work I agree it should be major.

PR #135, opened 2020-04-17
Allows all records for addresses

Justintime50:
The Node client library does not allow users to retrieve all `addresses` but the other CL's do. I've removed the `NotImplemented` code to allow users to retrieve all records as 1) the API supports this functionality and 2) the other CL's allow for these methods on these objects.


Justintime50:
I'm realizing I've compiled the library in this PR, let me know if you'd like me to remove those and just have the code changes here and then compile/update version in a following PR.

PR #166, opened 2021-05-14
feat: adds smartrate functionality

Justintime50:
Adds the new `/smartrate` endpoint/method on the `Shipment` object.

Simply call `getSmartrates` on the shipment object and you'll be returned a list of smartrates like so:

```json
[
        {
            "carrier": "USPS",
            "carrier_account_id": "ca_123...",
            "created_at": "2021-05-04T21:15:59Z",
            "currency": "USD",
            "delivery_date": null,
            "delivery_date_guaranteed": true,
            "delivery_days": null,
            "est_delivery_days": null,
            "id": "rate_123...",
            "list_currency": null,
            "list_rate": null,
            "mode": "test",
            "object": "Rate",
            "rate": 0.01,
            "retail_currency": null,
            "retail_rate": null,
            "service": "First",
            "shipment_id": "shp_123",
            "time_in_transit": {
                "percentile_50": 2,
                "percentile_75": 3,
                "percentile_85": 4,
                "percentile_90": 4,
                "percentile_95": 4,
                "percentile_97": 4,
                "percentile_99": 4
            },
            "updated_at": "2021-05-04T21:15:59Z"
        },
    ]
```

A full code example would look like this:

```python
api.Shipment.retrieve('shp_123...').then(shipment => {
  shipment.getSmartrates().then(console.log);
});
```


Justintime50:
Let me know if we want to generate the new build assets with the version bump commit instead and I can pull those back out of this one.

src/resources/base.js
```diff
@@ -143,6 +144,9 @@ export default api => (
           res = await api[method](url);
         }
 
+        if (pathsWithoutEasyPostObjects.includes(path)) {
```
Justintime50, line 147: # Reviewer Note

This was done because we don't want to append the `smartrates` results to the `Shipment` object, we only want the `results`. This list can be extensible if we had other methods that needed to take advantage in the future.

test/resources/shipment.js
```diff
@@ -241,6 +241,42 @@ describe('Shipment Resource', () => {
     });
   });
 
+  describe('retrieving smartrates', () => {
```
Justintime50, line 244: # Reviewer Note

These tests don't really do much for us but I followed the same pattern as other tests.

I also end-to-end tested to ensure we get back the expected results when requesting smartrates and that things like creating a shipment continued to behave as expected.

src/resources/shipment.js
```diff
@@ -110,6 +110,14 @@ export default api => (
       return this.rpc('rates', undefined, undefined, 'get');
     }
 
+    async getSmartrates() {
+      this.verifyParameters({
+        this: ['id'],
+      });
+
+      return this.rpc('smartrate', undefined, undefined, 'get');
```
jontsai, line 118: We'll probably want to do something like this to future-proof our API:

```suggestion
      const response = this.rpc('smartrate', undefined, undefined, 'get');
      const smartRates = (typeof response.result === 'Array' ? response.result : response) || [];
      return smartRates;
```

jontsai (changes requested):

ajnassar (dismissed):
LGTM

src/resources/base.js
```diff
@@ -143,6 +143,10 @@ export default api => (
           res = await api[method](url);
         }
 
+        if (path === 'smartrate') {
```
jontsai, line 146: Why here instead of in the `getSmartrates` function?

src/resources/shipment.js
jontsai: Do we want to do a protected get?

src/resources/shipment.js
```diff
@@ -110,6 +110,14 @@ export default api => (
       return this.rpc('rates', undefined, undefined, 'get');
     }
 
+    async getSmartrates() {
```
jontsai, line 113: Should we camel case `getSmartRates`?

justinwaite:
Following... New customer here, excited to get this integrated 👋🏻

src/resources/base.js
Justintime50: The function doesn't yet have the response available to grab specific keys from. It's at this point that we can strip away the key before sending it to the user.

src/resources/shipment.js
Justintime50: I didn't in the Golang library, I suppose if we are making a decision it should be done now and followed everywhere before we get too deep. I'm not super set glued to either one.

Justintime50:
> LGTM

@ajnassar looks like you left a comment instead of also approving. Could you re-review and select `approve` this time? Thanks!

ajnassar (approved):

Justintime50:
@jontsai can I get a re-review?

PR #172, opened 2021-06-08
fix: adds declaration propType to customsInfo

Justintime50:
Adds missing `declaration` to propType of customsInfo.


Justintime50:
Bad branch renaming, closing in favor of #173

PR #179, opened 2021-06-11
fix: changes usps_zone from integer to number type

Justintime50:
Apparently the `propType` of an `integer` is supposed to be a `number` here: https://www.npmjs.com/package/prop-types

Fixes the `usps_zone` field type.


jontsai (approved):
LGTM

ajnassar (approved):

PR #186, opened 2021-10-01
fix: json encode bodies instead of urlencoding

Justintime50:
We had both `json` and `urlencoding` present here. The `json` encoding "default" would get overridden by the `urlencoding` which made adding things like `tax_identifiers` impossible to add. Now we encode with JSON by default as this is the direction we want to take our API and client libraries.

I added `tax_identifiers` separately and got it working with this fix.

### DIFF
--- a/src/easypost.js
+++ b/src/easypost.js
@@ -41,8 +41,9 @@ export const UA_INFO = {
 export const EASYPOST_UA_HEADER = 'X-EasyPost-Client-User-Agent';
 
 export const DEFAULT_HEADERS = {
+  Accept: 'application/json',
   'Accept-Encoding': 'gzip,deflate,sdch,br',
-  'Content-Type': 'application/x-www-form-urlencoded',
+  'Content-Type': 'application/json',
   'User-Agent': `EasyPost/v2 NodejsClient/${pkg.version}`,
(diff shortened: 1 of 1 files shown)


test/easypost.js
```diff
@@ -126,13 +126,6 @@ describe('Base API object', () => {
       });
     });
 
-    it('requets json', done => {
```
fiendish, line 129: What was this? (and why does it say 'requets'?)

Justintime50: Someone left a typo 😉 

It was a test asserting we passed `Accept json` into the headers, which we now do elsewhere and so this test is now dead.

ajnassar (approved):

test/easypost.js
```diff
@@ -178,7 +171,7 @@ describe('Base API object', () => {
     it('handles request failures with a RequestError', done => {
       api.agent = superagentStub(true);
 
-      api.request('', METHODS.POST).then(() => {}, err => {
+      api.request('', METHODS.POST).then(() => { }, err => {
```
jontsai, line 174: # Reviewer Note

Fixes spacing for linter

jontsai (approved):
LGTM

PR #199, opened 2022-01-24
feat: adds the Node version to the user-agent header

Justintime50:
Adds the Node version to the user-agent header.

Tested via the repl:
```bash
jhammond@easypost-node $ node
Welcome to Node.js v14.18.2.
Type ".help" for more information.
> process.versions.node
'14.18.2'
```


nwithan8 (approved):
LGTM, seems like there's long-term support for this method:
https://nodejs.org/api/process.html#process_process_version

PR #201, opened 2022-02-02
Create list tracker

jchen293:
## What?
Add functionality to create trackers in bulk for our node library.
## Why?
Currently, our node library is missing one of the core features.
## How?
Add CreateList method in Tracker class.
## Testing?
Add a unit test in tracker.js test file.

### DIFF
--- a/src/resources/tracker.js
+++ b/src/resources/tracker.js
@@ -36,7 +36,7 @@ export default api => (
     createList(param) {
       const newParam = new Map();
       newParam.set('trackers', param);
-      this.rpc('create_list1', newParam);
+      this.rpc('create_list', newParam);
       return true;
     }
   }
--- a/src/resources/tracker.js
+++ b/src/resources/tracker.js
@@ -34,11 +34,11 @@ export default api => (
     }
 
     static async createList(params = {}) {
-      const newParam = { trackers: params };
+      const newParams = { trackers: params };
       try {
         const url = 'trackers/create_list';
-        await api.post(url, { body: newParam });
-        // This endpoint does not return anything so true is return here
+        await api.post(url, { body: newParams });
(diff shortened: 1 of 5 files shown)


test/resources/tracker.js
```diff
@@ -31,4 +31,19 @@ describe('Tracker Resource', () => {
       expect(err).to.be.an.instanceOf(NotImplementedError);
     });
   });
+
+  describe('Trackers in bulk', () => {
```
Justintime50, line 35: Did you actually run this unit test? Our Node library is a bit different in the fact that the unit test suite makes 0 API calls and everything is mocked (we'll be fixing this shortly). I don't know that we had a way built out for the tests to accept an API key to actually make this call here.

jchen293: Yes, I just E2E tested it.

src/resources/tracker.js
```diff
@@ -32,5 +32,17 @@ export default api => (
     static delete() {
       return this.notImplemented('delete');
     }
+
+    static async createList(params = {}) {
+      const newParam = { trackers: params };
```
Justintime50, line 37: `newParam` should be plural

src/resources/tracker.js
```diff
@@ -32,5 +32,17 @@ export default api => (
...
+      try {
+        const url = 'trackers/create_list';
+        await api.post(url, { body: newParam });
+        // This endpoint does not return anything so true is return here
```
Justintime50, line 41:
```suggestion
        // This endpoint does not return anything so true is returned here
```

Justintime50 (changes requested):

Justintime50 (approved):

nwithan8 (approved):
LGTM

PR #202, opened 2022-02-04
add support for update_brand.

jchen293:
# What
Added support for updating the user brand in the Node Library, updated the changelog, and add code example.

# Why
This is one of the missing core features.

# How
Added **updateBrand()** method in User class and add a new file for a code example. Created a new class for Brand.

# Testing
E2E local testing, no unit test was written because there is my personal information in the cassette.

### DIFF
--- a/src/resources/brand.js
+++ b/src/resources/brand.js
@@ -1,0 +1,22 @@
+import T from 'proptypes';
+import base from './base';
+
+
+export const propTypes = {
+  background_color: T.string,
+  color: T.string,
+  logo: T.string,
+  logo_href: T.string,
+  ad: T.string,
+  ad_href: T.string,
+  name: T.array,
+  user_id: T.array,
+  theme: T.array,
+};
+
+
+export default api => (
+  class Brand extends base(api) {
+
+  }
+);
--- a/src/resources/brand.js
+++ b/src/resources/brand.js
@@ -9,9 +9,9 @@ export const propTypes = {
   logo_href: T.string,
   ad: T.string,
   ad_href: T.string,
-  name: T.array,
-  user_id: T.array,
-  theme: T.array,
+  name: T.string,
+  user_id: T.string,
+  theme: T.string,
 };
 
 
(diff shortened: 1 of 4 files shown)


src/resources/brand.js
```diff
@@ -0,0 +1,22 @@
...
+export const propTypes = {
+  background_color: T.string,
+  color: T.string,
+  logo: T.string,
+  logo_href: T.string,
+  ad: T.string,
+  ad_href: T.string,
+  name: T.array,
+  user_id: T.array,
+  theme: T.array,
```
Justintime50, line 14: These fields need to be strings.

src/resources/user.js
```diff
@@ -42,5 +42,19 @@ export default api => (
     static all() {
       return this.notImplemented('all');
     }
+
+    async updateBrand(param, id) {
```
Justintime50, line 46: 1. `param` needs to be plural - `params`
2. Why are we passing in an `id` parameter here? We should be getting the instance ID of the user.

Justintime50 (changes requested):

src/resources/user.js
jchen293: Corrected

src/resources/brand.js
jchen293: Corrected

Justintime50 (approved):

src/resources/user.js
```diff
@@ -42,5 +42,19 @@ export default api => (
...
       return this.notImplemented('all');
     }
+
+    async updateBrand(params) {
+      try {
+        const newParams = { brand: params };
+        const urlPrefix = this.constructor._url;
+        const userData = this.constructor.wrapJSON(this.toJSON());
+        const endpoint = 'brand';
+        const url = `${urlPrefix}/${userData.user.id}/${endpoint}`;
```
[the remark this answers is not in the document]
jchen293: The eslint suggests using `const` since the variable is never reassigned. I have removed `urlPrefix` and `endpoint` variables.

nwithan8 (approved):

Justintime50 (approved):

PR #204, opened 2022-02-08
remove retrieveRate() method and its unit tests.

jchen293:
This PR removes **retrieveRate()** in the shipment class and its belonging unit tests because the shipment object already has rates, which makes this method unnecessary. If you need to get new rates for a shipment, please use **regenerateRate()** method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,7 @@
 * Added the Nodejs version in use to the User-Agent header on requests
 * Added support for create list of trackers
 * Added support for update brand for user
-* Removed `RetrieveRate()` method because the shipment object already
-* has rates. If you need to get new rates for a shipment, please use
-* regenerateRates() method instead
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
 * Added the Nodejs version in use to the User-Agent header on requests
 * Added support for create list of trackers
 * Added support for update brand for user
+* Removed `RetrieveRate()` method because the shipment object already
+* has rates. If you need to get new rates for a shipment, please use
+* regenerateRates() method instead
```
Justintime50, line 12: Let's slow down and really read through things before we commit. For instance, the method name in this library is `retrieveRates`.

Also, each of these `*` is a bullet point in markdown and not a comment. We don't break up long lines in markdown because 1) we don't have a linter for markdown and don't really care about the line length and 2) this would add 3 bullet points for one thought. Please make this line a single line.

Justintime50 (changes requested):

nwithan8 (approved):

PR #214, opened 2022-02-25
chore: drops support for Node < 10

Justintime50:
Drops support for Node < 10 and removes the build matrix for those versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,8 +11,7 @@ EasyPost is a simple shipping API. You can sign up for an account at https://eas
 npm install --save @easypost/api
 ```
 
-**Note:** if you are using a version of Node less than 6.9, you will need to install and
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -11,31 +11,8 @@ EasyPost is a simple shipping API. You can sign up for an account at https://eas
 npm install --save @easypost/api
 ```
 
-**Note:** if you are using a version of Node less than 6.9, you will need to install and
```
jchen293, line 14: Do we want to keep this? What if the user is using the older version of our Node CL?

Justintime50: hmm, yeah we probably just need to clarify this to say if you're using our library prior to x version. I'll fix this.

jchen293 (approved):

nwithan8 (approved):

PR #217, opened 2022-03-03
chore: bump all prod and dev dependencies

Justintime50:
Bumps all dependencies to the most recent compatible versions that still support Node 10, the library is now buildable on Node 16+.

With the adoption of `eslint 5` and v2 of our `eslint-config-easypost-base`, I had to rename the files from camelCase to snake_case to get linting to pass. There is no logic change in this PR aside from adjusting references to files. The only *real* change here is adding Node 16 to CI and overhauling the `package.json` file.

The following dependencies were removed:
- Removed `@babel/plugin-proposal-nullish-coalescing-operator` and `@babel/plugin-proposal-optional-chaining` as they were included in `@easypost/build` but we weren’t using them in this project
- Removed `eslint-config-airbnb` because it’s included in `eslint-config-easypost-base` v2
- Removed `babel-eslint` because it’s included in `eslint-config-easypost-base` v2
- Upgraded `eslint` from v4 to v5 so we can adopt our newer config easily but left `eslint` on v5 which is pretty dated but all our configs depend on that version

FUTURE:
- May be able to remove `sinon`, `sinon-chai`, and `chai-as-promised` when we stop mocking everything and use VCR instead as we could probably get away with the basic `chai` package alone


test/resources/api_key.js
```diff
@@ -17,26 +17,26 @@ describe('ApiKey Resource', () => {
     expect(apiKey).to.be.a('function');
   });
 
-  it('throws on delete', () => ApiKey.delete('id').then(() => {}, err => {
+  it('throws on delete', () => ApiKey.delete('id').then(() => { }, err => {
     expect(err).to.be.an.instanceOf(NotImplementedError);
   }));
```
hagemt, line 22: A small problem I see (which this PR doesn't need to address) is:

Won't test like this one pass if the Promise **doesn't** reject, too? Probably not urgent, but seems fishy to me.

hagemt (approved):

test/resources/api_key.js
Justintime50: The whole test suite in this library is awful. We have an internal ticket to overhaul it completely which I'll be doing next week (most of the issues with testing should get addressed there). Thanks for calling this out!

jchen293 (approved):

PR #218, opened 2022-03-04
chore: adds prettier and formats the project

Justintime50:
Adds `prettier` and formats the entire project. Only rules were:

- Line length of 100 (to match eslint config)
- Trailing comma always (internal style preference)
- Keep single quotes (style already used throughout project)

Also moved the built assets to a `dist` folder for easier cleaning/ignoring

### DIFF
--- a/examples/shipments/buy-shipment.js
+++ b/examples/shipments/buy-shipment.js
@@ -1,9 +1,10 @@
 const Easypost = require('@easypost/api');
 
-
 const api = new Easypost(process.env.API_KEY);
 
-api.Shipment.retrieve('shp_...').then(s => {
-  s.buy(s.lowestRate()).then(console.log).catch(console.log);
-  // s.buy('rate_...').then(console.log).catch(console.log);
-}).catch(console.log);
+api.Shipment.retrieve('shp_...')
+  .then((s) => {
+    s.buy(s.lowestRate()).then(console.log).catch(console.log);
+    // s.buy('rate_...').then(console.log).catch(console.log);
+  })
+  .catch(console.log);
--- a/.babelrc
+++ b/.babelrc
@@ -1,58 +1,56 @@
 {
-    "presets": [
-        [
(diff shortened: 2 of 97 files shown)


examples/shipments/buy-shipment.js
```diff
@@ -1,9 +1,10 @@
...
 const api = new Easypost(process.env.API_KEY);
 
-api.Shipment.retrieve('shp_...').then(s => {
-  s.buy(s.lowestRate()).then(console.log).catch(console.log);
-  // s.buy('rate_...').then(console.log).catch(console.log);
-}).catch(console.log);
+api.Shipment.retrieve('shp_...')
+  .then((s) => {
+    s.buy(s.lowestRate()).then(console.log).catch(console.log);
+    // s.buy('rate_...').then(console.log).catch(console.log);
```
jchen293, line 8: can we remove this?

nwithan8 (approved):
Won't block, but technically, JSON is supposed to explicitly use double quotes per the spec: https://www.ecma-international.org/publications-and-standards/standards/ecma-404/

I know it doesn't really matter functionality-wise, and the linter has an option to enable/disable this, but some IDEs might complain about it.

jchen293 (approved):

examples/shipments/buy-shipment.js
Justintime50: Can you be more explicit about what you mean by `this`? It's an example showing you can either buy with the lowest rate or by passing in a rate ID. I won't be removing it as I believe it's useful and also not related to the changes made here, though we should go through and update all the examples separately (we have an internal task to do so).

jchen293: Sorry, I meant the comment in line 8: `// s.buy('rate_...').then(console.log).catch(console.log);`

ldewald (approved):
Yay formatting

PR #219, opened 2022-03-07
chore: bump easypost/eslint-config-easypost-base

Justintime50:
Bumps `easypost/eslint-config-easypost-base` to correct deprecation warning and add support for eslint 5 which is in use. Additionally, with the bump of the style comes the need for the `eslint-config-prettier` dependency so that conflicting rules between `prettier` and `eslint` get ignored.


jchen293 (approved):

test/.eslintrc
```diff
@@ -1,7 +1,15 @@
...
+    "@easypost/eslint-config-easypost-base",
+    "prettier",
+    "prettier/@typescript-eslint",
+    "prettier/babel",
+    "prettier/flowtype",
+    "prettier/react",
+    "prettier/standard",
+    "prettier/unicorn",
+    "prettier/vue"
+  ],
```
jchen293, line 13: I think we have an extra line here

jchen293 (approved):

test/.eslintrc
Justintime50: Yeah, each of these sections is separated by a line.

PR #220, opened 2022-03-08
fix: customs_item.value propType

Justintime50:
Corrects the `customs_item.value` propType from string to number. This should have always been number, because the Javascript spec defines a number as a whole number or a decimal. Our docs have this field as a float so this now more clearly defines what is acceptable in this field:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Add `retrieveMe()` Convenience Function that allow users to retrieve without specifying an ID.
 - Removed `enable()` and `disable()` methods in the apiKey class. Please use this functionality through EasyPost website
 - Removed `options.useCookie` from the parameter since it's deprecated in v4.0. Please use `options.useProxy` instead.
+- Fixes `customs_item.value` to ve a `number` instead of a `string`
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 - Add `retrieveMe()` Convenience Function that allow users to retrieve without specifying an ID.
 - Removed `enable()` and `disable()` methods in the apiKey class. Please use this functionality through EasyPost website
 - Removed `options.useCookie` from the parameter since it's deprecated in v4.0. Please use `options.useProxy` instead.
+- Fixes `customs_item.value` to ve a `number` instead of a `string`
```
jchen293, line 20:
```suggestion
- Fixes `customs_item.value` to be a `number` instead of a `string`
```

jchen293 (approved):

CHANGELOG.md
Justintime50: Oof, thanks. I'll fix this when we release it and I rework the CHANGELOG so I don't dismiss the review.

PR #223, opened 2022-03-12
Overhauls the Test Suite to Make Real API Calls (With VCR)

Justintime50:
The library now makes real API calls using VCR to record the requests and responses as we've done with other libs and targeted coverage at ~70% - I was able to get it up to 87% by testing everything reasonable on this first pass. This provides much better visibility into how the library behaves over the previous mocking strategy. 

Various bugs were fixed and unused code removed due to this effort.

Screen Shot 2022-03-16 at 3 13 01 PM


nwithan8 (approved):
LGFM

test/helpers/fixture.js
```diff
@@ -0,0 +1,195 @@
+export default class Fixture {
+  static pageSize() {
+    return 5;
+  }
+
+  static uspsCarrierAccountId() {
+    return 'ca_5ba7ca3632c54adeb17ad4bcac13c890';
+  }
```
jchen293, line 8: Add missing comments here
```suggestion
  // We keep the page_size of retrieving `all` records small so cassettes stay small
  static pageSize() {
    return 5;
  }

  // This is the carrier account ID for the default USPS account that comes by default. All tests should use this carrier account
  static uspsCarrierAccountId() {
    return 'ca_5ba7ca3632c54adeb17ad4bcac13c890';
  }
```

test/resources/insurance.test.js
```diff
@@ -0,0 +1,67 @@
...
+    const shipment = await new this.easypost.Shipment(Fixture.oneCallBuyShipment()).save();
+
+    const insuranceData = Fixture.basicInsurance();
+    insuranceData.tracking_code = shipment.tracking_code;
+
+    const insurance = await new this.easypost.Insurance(insuranceData).save();
+
+    expect(insurance).to.be.an.instanceOf(this.easypost.Insurance);
+    expect(insurance.id).to.match(/^ins_/);
+  });
```
jchen293, line 30: you should add assertion to compare the insurance amount here

test/resources/scan_form.test.js
```diff
@@ -0,0 +1,66 @@
...
+
+    const scanform = await new this.easypost.ScanForm({
+      shipments: [shipment],
+    }).save();
+
+    const retrievedScanform = await this.easypost.ScanForm.retrieve(scanform.id);
+
+    expect(retrievedScanform).to.be.an.instanceOf(this.easypost.ScanForm);
+    delete scanform.shipments; // This lib returns the shipments key as part of the response, remove it for comparison
+    expect(retrievedScanform.id).to.equal(scanform.id);
```
jchen293, line 42: Can we compare the objects here?

jchen293 (changes requested):
Nice work, I didn't see the unit tests for refund, are we not testing it in node.js?

test/resources/scan_form.test.js
Justintime50: Ah yes, this was leftover from some bugs I had to deal with. Fixed

jchen293 (approved):

nwithan8 (approved):
LGFM

jchen293 (approved):

PR #233, opened 2022-04-07
add refund class.

jchen293:
Our API allows you to either refund a shipment from the `shipment` object or from the `Refund` class; however, the `refund` class is currently missing from the Node library.

This PR adds the missing `Refund` class and unit tests to make sure they are working as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - Add support for `columns` and `additional_columns` on Report creation
 - Adds `address.verifyAddress()` and `Address.createAndVerify()` functions
 - Adds `Batch.createAndBuy()` function
+- Adds `Refund` class which has `Refund.save()`, `Refund.all()`, and `Refund.retrieve()`
 
 ## 4.0.0 2021-10-06
 
--- a/src/resources/refund.js
+++ b/src/resources/refund.js
@@ -1,0 +1,44 @@
+import T from 'proptypes';
+import base from './base';
+
+export const propTypes = {
+  id: T.string,
+  tracking_code: T.string,
+  confirmation_number: T.string,
+  status: T.string,
+  carrier: T.string,
+  shipment_id: T.string,
+};
+
+export default (api) =>
+  class Refund extends base(api) {
+    static _name = 'Refund';
+
+    static _url = 'refunds';
+
+    static key = 'refund';
+
+    static propTypes = propTypes;
+
+    /**
+     * create a refund
+     * @returns {Promise<never>}
+     */
+    static async save(params = {}) {
+      const newParams = { refund: params };
+      try {
+        const response = await api.post(this._url, { body: newParams });
+        return response.body;
+      } catch (e) {
+        return Promise.reject(e);
+      }
+    }
+
+    /**
+     * delete not implemented
+     * @returns {Promise<never>}
+     */
+    static delete() {
+      return this.notImplemented('delete');
+    }
+  };
--- a/test/resources/refund.test.js
+++ b/test/resources/refund.test.js
@@ -1,0 +1,65 @@
+/* eslint-disable func-names */
+import { expect } from 'chai';
+import * as setupPolly from '../setup_polly';
+import EasyPost from '../../src/easypost';
+import Fixture from '../helpers/fixture';
+import NotImplementedError from '../../src/errors/not_implemented';
+
+describe('Refund Resource', function () {
+  setupPolly.startPolly();
+
+  before(function () {
+    this.easypost = new EasyPost(process.env.EASYPOST_TEST_API_KEY);
+  });
+
+  beforeEach(function () {
+    const { server } = this.polly;
+    setupPolly.stripCreds(server);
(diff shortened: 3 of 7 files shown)


src/resources/refund.js
```diff
@@ -0,0 +1,44 @@
...
+
+    static key = 'refund';
+
+    static propTypes = propTypes;
+
+    /**
+     * create a refund
+     * @returns {Promise<never>}
+     */
+    static async save(params = {}) {
```
Justintime50, line 27: Do we need a custom overridden function here to save or can we use the `base` save function instead?

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 - Fixes `customs_item.value` to ve a `number` instead of a `string`
 - Add support for `columns` and `additional_columns` on Report creation
 - Adds `address.verifyAddress()` and `Address.createAndVerify()` functions
+- Adds `Refund` class
```
Justintime50, line 26: Can we instead add what functions of the Refund class were added? Simply stating `Refund` got added doesn't help users know what is now accessible.

test/resources/refund.test.js
```diff
@@ -0,0 +1,65 @@
...
+  before(function () {
+    this.easypost = new EasyPost(process.env.EASYPOST_TEST_API_KEY);
+  });
+
+  beforeEach(function () {
+    const { server } = this.polly;
+    setupPolly.stripCreds(server);
+  });
+
+  it('create a refund', async function () {
```
Justintime50, line 20:
```suggestion
  it('creates a refund', async function () {
```

test/resources/refund.test.js
```diff
@@ -0,0 +1,65 @@
...
+    const shipment = await new this.easypost.Shipment(Fixture.oneCallBuyShipment()).save();
+
+    const retrievedShipment = await this.easypost.Shipment.retrieve(shipment.id);
```
Justintime50, line 23: Please add this comment above:

`We need to retrieve the shipment so that the tracking_code has time to populate`

test/resources/refund.test.js
```diff
@@ -0,0 +1,65 @@
...
+
+    const refundData = {
+      carrier: Fixture.usps(),
+      tracking_codes: retrievedShipment.tracking_code,
+    };
+
+    const refunds = await this.easypost.Refund.save(refundData);
+
+    expect(refunds).to.be.an.instanceOf(Array);
```
Justintime50, line 32: Let's not assert this, no need (really it's unfortunate that this endpoint returns an array).

test/resources/refund.test.js
```diff
@@ -0,0 +1,65 @@
...
+    expect(refunds.has_more).to.not.be.null;
+    refundsArray.forEach((address) => {
+      expect(address).to.be.an.instanceOf(this.easypost.Refund);
+    });
+  });
+
+  it('retrieve a refund', async function () {
+    const refunds = await this.easypost.Refund.all({ page_size: Fixture.pageSize() });
+
+    const retrieveRefund = await this.easypost.Refund.retrieve(refunds.refunds[0].id);
```
Justintime50, line 52: Please add this comment:

`We need to retrieve the shipment so that the tracking_code has time to populate`

src/resources/refund.js
jchen293: The `base` save function is non-static and we need to have a wrappedParams for the `refund` parameter.

test/resources/refund.test.js
jchen293: I can remove it, but other libs have this assertion.

test/resources/refund.test.js
jchen293: I don't think this comment is needed because we are not creating a new shipment. Instead, we are retrieving all refunds and then using the first refund in the array.

Justintime50: Ah I see, you're right! 😃

test/resources/refund.test.js
Justintime50: Eh, sure - either way

Justintime50 (approved):

src/resources/refund.js
```diff
@@ -0,0 +1,44 @@
...
+        return Promise.reject(e);
+      }
+    }
+
+    /**
+     * delete not implemented
+     * @returns {Promise<never>}
+     */
+    static delete() {
+      return this.notImplemented('delete');
```
nwithan8, line 42: Why is this not implemented?

nwithan8 (changes requested):

src/resources/refund.js
jchen293: The `refund` class should not have `delete()` method. The parent class `base` has the `delete()` method, so we need to override it.

Justintime50: Yeah, this is a standard pattern in this lib.

Justintime50 (approved):

nwithan8 (approved):

PR #242, opened 2022-04-25
White label

nwithan8:
# Description

- Add Beta API class for accessing beta features
- Add `Referral` class under beta for creating, retrieving, updating, and adding credit cards for referral users
- Add test suite for referrals

# Testing

Two new recorded tests (retrieve, update).
Two E2E tests (create, add credit card) due to sensitive information.

# Pull Request Type


src/beta/resources/referral.js
```diff
@@ -0,0 +1,117 @@
...
+     * @param {string} creditCardDetails.expirationMonth  - Credit card expiration month
+     * @param {string} creditCardDetails.expirationYear  - Credit card expiration year
+     * @param {string} creditCardDetails.cvc  - Credit card CVC
+     * @param {string} primaryOrSecondary  - Whether to add the card as the primary or secondary card (defaults to primary)
+     * @returns {Promise<never>}
+     */
+    async addCreditCard(creditCardDetails = {}, primaryOrSecondary = "primary") {
+      const stripeCreditCardId = await Referral.#sendCardDetailsToStripe(creditCardDetails);
+      await Referral.#sendCardDetailsToEasyPost(stripeCreditCardId, primaryOrSecondary);
+      // nothing to return. If an error wasn't thrown, then the card was added successfully
```
Justintime50, line 65: This should return a response form EasyPost (EasyPost card object)

nwithan8: I'm actually noticing that a lot of things in the Node library don't return as a true class.
For example, we don't actually return a `Brand` object when updating a user's brand; we return the body of the request, which is JSON (which acts like an object because, JSON in JavaScript...), but we're not actually saying "create a Brand object out of this data"

That said, I attempted to do so with a couple things, including `Brand` and here, taking the data that's returned and build a `PaymentMethod` object out of it, but I'm getting complaints about constructors. Which is probably the reason why it is the way it is.

So anyway, this function now returns the body of the final request, which is the JSON representation of a payment method, but isn't technically a `PaymentMethod` object.

Justintime50: I ran into this same problem when adding a few rate/brand unit tests when we overhauled the test suite. Can you create an internal task with some details to look into this with the Node lib? We either need to answer why it's doing this or fix it. Good callout.

nwithan8: As it stands, it seems to be an issue whenever a class is attempting to return a different type of class other than itself.

For example, most functions in the `User` class return a `User` object, by taking the data from the API response and mapping its properties via [this base `create` function](https://github.com/EasyPost/easypost-node/blob/eb4184d0b33fdf98f90ef56bb6ce59e1c23292f2/src/resources/base.js#L180). But `updateBrand` in the `User` class returns a `Brand` object, which throws the constructor error whenever we try to build a `Brand` object.

I imagine it's due to how the `base` class is designed for object creation, as well as perhaps the module exporting setup.

src/beta/resources/referral.js
```diff
@@ -0,0 +1,137 @@
+/* eslint-disable no-dupe-class-members,no-unused-vars,class-methods-use-this */
```
jchen293, line 1: Remove this comment

src/beta/resources/referral.js
```diff
@@ -0,0 +1,137 @@
...
+      return super.save();
+    }
+
+    /**
+     * Update the referral's email address.
+     * @param {string} referralUserId - The referral user's ID.
+     * @param {string} email - The new email address.
+     * @returns {Promise<boolean>}
+     */
+    static async updateEmail(referralUserId, email) {
```
jchen293, line 48: This function should be called `updateReferral` according to our internal doc

jchen293, line 48: This function should be called `updateReferral` according to our internal doc

src/beta/resources/referral.js
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Add a credit card to the referral's account
+     * @param {string} referralApiKey - The referral user's production API key.
+     * @param {object} creditCardDetails  - Credit card details:
+     * @param {string} creditCardDetails.number  - Credit card number
+     * @param {string} creditCardDetails.expirationMonth  - Credit card expiration month
+     * @param {string} creditCardDetails.expirationYear  - Credit card expiration year
+     * @param {string} creditCardDetails.cvc  - Credit card CVC
+     * @param {string} primaryOrSecondary  - Whether to add the card as the primary or secondary card (defaults to primary)
```
jchen293, line 66:
```suggestion
     * @param {object} creditCardDetails - Credit card details:
     * @param {string} creditCardDetails.number - Credit card number
     * @param {string} creditCardDetails.expirationMonth - Credit card expiration month
     * @param {string} creditCardDetails.expirationYear - Credit card expiration year
     * @param {string} creditCardDetails.cvc - Credit card CVC
     * @param {string} primaryOrSecondary - Whether to add the card as the primary or secondary card (defaults to primary)
```

Should this be single spaced

src/beta/resources/referral.js
```diff
@@ -0,0 +1,137 @@
...
+      const stripeCreditCardId = await Referral.#sendCardDetailsToStripe(creditCardDetails);
+      const paymentMethod = await Referral.#sendCardDetailsToEasyPost(
+        referralApiKey,
+        stripeCreditCardId,
+        primaryOrSecondary,
+      );
+      return paymentMethod;
+    }
+
+    static async #getStripeKey() {
```
jchen293, line 83: The key is from EasyPost, the function name should be renamed.

.gitignore
```diff
@@ -14,3 +14,5 @@ dist
 coverage
 
 build.tar.gz
+/test/cassettes/Referral-Resource_2042494388/add-a-referral-user-credit-card_1978527022/recording.har
+/test/cassettes/Referral-Resource_2042494388/creates-a-referral-user_159009729/recording.har
```
Justintime50, line 18: Please remove these. Pollyjs gives unique IDs to each cassette so this is an ineffective way to protect against accidental inclusion of a specific cassette.

test/resources/referral.test.js
```diff
@@ -0,0 +1,95 @@
...
+    const { server } = this.polly;
+    setupPolly.stripCreds(server);
+  });
+
+  async function createTestReferral() {
+    // Note: this returns API keys in plaintext, so do not record a test using this function
+    const referral = await new this.easypost.Referral({
+      name: 'Test Referral',
+      email: 'me@email.com',
+      phone_number: '1111111111', // phone_number will be copied to phone behind the scenes
```
jchen293, line 24: I think the API endpoint is returning `phone_number` instead of `phone` now, you can remove the comment.

src/beta/resources/referral.js
```diff
@@ -0,0 +1,137 @@
...
+
+import T from 'proptypes';
+import base from '../../resources/base';
+import { propTypes as userPropTypes } from '../../resources/user';
+import RequestError from '../../errors/request';
+import API from '../../easypost';
+
+export const propTypes = Object.assign({}, userPropTypes, {
+  name: T.string.isRequired, // mark as required
+  email: T.string.isRequired, // mark as required
```
Justintime50, line 12: The comments stating it's required directly after the property `isRequired` doesn't help and instead muddies up the code. Let's remove these two comments.

src/beta/resources/referral.js
```diff
@@ -0,0 +1,137 @@
...
+  class Referral extends base(api) {
+    static _name = 'Referral';
+
+    static _url = 'referral_customers';
+
+    static key = 'user';
+
+    static propTypes = propTypes;
+
+    static #stripeCreditCardUrl = 'https://api.stripe.com/v1/tokens';
```
Justintime50, line 26: My understanding is that the "private function/vars" addition of `#` was only added in Node 12+ and we support Node 10. We'll need to look at a pollyfill to allow this to get backported to Node 10. It may not have failed here because we're simply replaying cassettes and you didn't record them on Node 10 (could very well fail at runtime): https://stackoverflow.com/a/56183753/6064135

This is true for all `#` usage in this PR

src/beta/resources/referral.js
```diff
@@ -0,0 +1,137 @@
...
+
+    static #getReferralApi(referralApiKey) {
```
Justintime50, line 28:
```suggestion
    static #getReferralApiKey(referralApiKey) {
```

src/beta/resources/referral.js
Justintime50, line 48: Hmm, do we know for sure they can only ever update the email via this endpoint? If they can update more, we should make it more generic. Anyway you could find out?

src/beta/resources/referral.js
Justintime50, line 66: Nit: There are extra spaces inbetween all these `-` that should just be a single space.

test/resources/referral.test.js
```diff
@@ -0,0 +1,95 @@
...
+      expirationYear: '2028',
+      cvv: '015',
+    };
+    const paymentMethod = await this.easypost.Referral.addCreditCard(
+      referralProdApiKey,
+      creditCardDetails,
+      'primary',
+    );
+
+    expect(paymentMethod).to.not.be.null;
```
jchen293, line 90: This can be removed

src/beta/resources/referral.js
```diff
@@ -0,0 +1,137 @@
...
+     * @param {string} creditCardDetails.number  - Credit card number
+     * @param {string} creditCardDetails.expirationMonth  - Credit card expiration month
+     * @param {string} creditCardDetails.expirationYear  - Credit card expiration year
+     * @param {string} creditCardDetails.cvc  - Credit card CVC
+     * @param {string} primaryOrSecondary  - Whether to add the card as the primary or secondary card (defaults to primary)
+     * @returns {Promise<never>}
+     */
+    static async addCreditCard(
+      referralApiKey,
+      creditCardDetails = {},
```
Justintime50, line 71: I thought we were going to have 4 separate parameters to this function instead of a dictionary for their card details?

src/beta/resources/referral.js
```diff
@@ -0,0 +1,137 @@
...
+import T from 'proptypes';
+import base from '../../resources/base';
+import { propTypes as userPropTypes } from '../../resources/user';
+import RequestError from '../../errors/request';
+import API from '../../easypost';
+
+export const propTypes = Object.assign({}, userPropTypes, {
+  name: T.string.isRequired, // mark as required
+  email: T.string.isRequired, // mark as required
+  phone: T.string.isRequired, // handle discrepancy between phone (create Referral) and phone_number (all User functions, retrieve Referral)
```
jchen293, line 13: remove the comment and rename `phone` to `phone_number`, the API endpoint is returning `phone_number` now

src/beta/resources/referral.js
Justintime50, line 83: This should probably be something like `getEasyPostStripeKey`? And then document what it does/why in a comment

src/beta/resources/referral.js
```diff
@@ -0,0 +1,137 @@
...
+      return API.copyApi(api, {
+        apiKey: referralApiKey,
+      });
+    }
+
+    async save() {
+      if (this.phone_number !== undefined) {
+        // handle discrepancy between phone (create Referral) and phone_number (all User functions, retrieve Referral)
+        this.phone = this.phone_number;
+      }
```
jchen293, line 38: This can be removed now

src/beta/resources/referral.js
```diff
@@ -0,0 +1,137 @@
...
+     * @param {string} creditCardDetails.cvc  - Credit card CVC
+     * @returns {Promise<string>} - Stripe credit card ID
+     */
+    static async #sendCardDetailsToStripe(creditCardDetails = {}) {
+      const { number, expirationMonth, expirationYear, cvc } = creditCardDetails;
+      const stripeKey = await Referral.#getStripeKey();
+
+      const req = superagent.post(this.#stripeCreditCardUrl).set({
+        Authorization: `Bearer ${stripeKey}`,
+        'Content-Type': 'application/x-www-form-urlencoded',
```
Justintime50, line 103: Can we include a comment here as to why we need to form encode it. Someone may try to swap it to JSON in the future without us calling it out here.

src/beta/resources/referral.js
```diff
@@ -0,0 +1,137 @@
...
+        },
+      });
+      try {
+        const res = await req;
+        return res.body.id;
+      } catch (e) {
+        if (e.response && e.response.body) {
+          throw new RequestError(e.response.body, this.#stripeCreditCardUrl);
+        }
+        throw e;
```
Justintime50, line 120: Let's be explicit over implicit. `res` should be `response`, `req` should be `request`, `e` should be `error`

src/beta/resources/referral.js
```diff
@@ -0,0 +1,137 @@
...
+
+    /*
+     * Send the credit card details to Stripe to get a Stripe credit card ID.
+     * @param {object} creditCardDetails  - Credit card details:
+     * @param {string} creditCardDetails.number  - Credit card number
+     * @param {string} creditCardDetails.expirationMonth  - Credit card expiration month
+     * @param {string} creditCardDetails.expirationYear  - Credit card expiration year
+     * @param {string} creditCardDetails.cvc  - Credit card CVC
+     * @returns {Promise<string>} - Stripe credit card ID
+     */
```
jchen293, line 96:
```suggestion
     * Send the credit card details to Stripe to get a Stripe credit card ID.
     * @param {object} creditCardDetails - Credit card details:
     * @param {string} creditCardDetails.number - Credit card number
     * @param {string} creditCardDetails.expirationMonth - Credit card expiration month
     * @param {string} creditCardDetails.expirationYear - Credit card expiration year
     * @param {string} creditCardDetails.cvc - Credit card CVC
     * @returns {Promise<string>} - Stripe credit card ID
     */
```
Singled spaced

src/easypost.js
```diff
@@ -126,6 +126,17 @@ export default class API {
     this.use(RESOURCES);
   }
 
+  static copyApi(api, options = {}) {
```
Justintime50, line 129: What is this? If we absolutely need it then we need to document it and potentially name it something more concrete.

src/easypost.js
```diff
@@ -201,7 +212,7 @@ export default class API {
       return res;
     } catch (e) {
       if (e.response && e.response.body) {
-        throw new RequestError(e.response.body, path);
```
Justintime50, line 204: Why are we changing the signature of errors here? This is a breaking change.

src/beta/resources/referral.js
```diff
@@ -0,0 +1,137 @@
...
+        'Content-Type': 'application/x-www-form-urlencoded',
+      });
+      req.send({
+        card: {
+          number,
+          exp_month: expirationMonth,
+          exp_year: expirationYear,
+          cvc,
+        },
+      });
```
jchen293, line 112: nit: need a new line before try-catch

src/errors/request.js
```diff
@@ -16,7 +16,10 @@ export default class RequestError extends FakeError {
       throw new Error('No url passed to RequestError');
     }
 
-    const message = createMessage(error.status || error.code, url);
+    const message = createMessage(
+      error.status || error.statusCode || error.code || error.body.code,
+      url,
+    );
```
Justintime50, line 22: We should not be altering the signature of errors. It's not related to this new feature and is a large breaking change.

src/beta/resources/referral.js
```diff
@@ -0,0 +1,137 @@
...
+      try {
+        const res = await req;
+        return res.body.id;
```
jchen293, line 115: nit: a new line before return statement

src/resources/base.js
```diff
@@ -93,6 +93,26 @@ export default (api) =>
...
+
+    static async rawPut(url, params = {}) {
+      const response = await api.put(url, params);
+      return response.body;
+    }
+
+    static async rawDelete(url, params = {}) {
+      const response = await api.delete(url, params);
+      return response.body;
+    }
```
Justintime50, line 114: None of this appears to be used? Why is it here?

test/errors/request.js
```diff
@@ -17,8 +17,9 @@ describe('Request Errors', () => {
   });
 
   it('maps code to status on sys errors', () => {
-    expect(new RequestError(sysError, url).status).to.equal(sysError.code);
-    expect(new RequestError(sysError, url).error.code).to.equal(sysError.code);
+    const requestError = new RequestError(sysError, url);
+    expect(requestError.status).to.equal(sysError.code);
+    expect(requestError.error.code).to.equal(sysError.code);
```
Justintime50, line 22: We should not be mixing adjustments to error handling in with a new feature.

src/beta/resources/referral.js
```diff
@@ -0,0 +1,137 @@
...
+        }
+        throw e;
+      }
+    }
+
+    /*
+     * Send the Stripe credit card ID to EasyPost to add the card to the user's account.
+     * @param {string} referralApiKey - The referral user's production API key.
+     * @param {string} stripeCreditCardId  - Stripe credit card ID
+     * @param {string} primaryOrSecondary  - Whether to add the card as the primary or secondary card (defaults to primary)
```
jchen293, line 128:
```suggestion
     * @param {string} referralApiKey - The referral user's production API key.
     * @param {string} stripeCreditCardId - Stripe credit card ID
     * @param {string} primaryOrSecondary - Whether to add the card as the primary or secondary card (defaults to primary)
```
singled spaced

test/resources/referral.test.js
```diff
@@ -0,0 +1,95 @@
...
+
+    const success = await this.easypost.Referral.updateEmail(referral.id, testEmail);
+
+    expect(success).to.eql(true);
+  });
+
+  it.skip('add a referral user credit card', async function () {
+    // skip due to exposed API keys, exposed tokens, and fake credit card details
+    // E2E test with real data
+    this.timeout(10000); // 10 seconds, since there are multiple API calls
```
Justintime50, line 72: The timeout of tests is already 10 seconds. As long as you do `npm run test`, this will be taken care of for you. Let's remove this.

jchen293, line 72: What is this for?

Justintime50 (changes requested):

src/beta/resources/referral.js
nwithan8: No, this gets an instance of the API with the Referral user's API key populated.

src/errors/request.js
nwithan8: Fixing a bug where a large portion of the time, we weren't extracting the error code correctly, so it would say "undefined" in the error message to the user.

src/beta/resources/referral.js
Justintime50: `api` was a terrible naming convention, whoever chose that years ago...

Let's document this then as to what it does. I still don't love the name. Could it be `getReferralApiClient` or something to that effect?

src/errors/request.js
Justintime50: Certainly that's not happening in all cases or we would have heard something by now from users. Can you provide examples?

src/beta/resources/referral.js
nwithan8: This is to stop the linter from aggressively complaining about linting inconsistencies that we don't care about.

src/beta/resources/referral.js
```diff
@@ -0,0 +1,162 @@
...
+      creditCardCVC,
+      primaryOrSecondary = 'primary',
+    ) {
+      const stripeCreditCardId = await sendCardDetailsToStripe(
+        api,
+        creditCardNumber,
+        creditCardExpirationMonth,
+        creditCardExpirationYear,
+        creditCardCVC,
+      );
```
jchen293, line 152: Let's add try-catch here in the event the Stripe endpoint doesn't return anything, so we can throw an explicit error to the users.

test/resources/referral.test.js
```diff
@@ -0,0 +1,98 @@
...
+
+    const success = await this.easypost.Referral.updateEmail(referral.id, testEmail);
+
+    expect(success).to.eql(true);
+  });
+
+  it.skip('add a referral user credit card', async function () {
+    // skip due to exposed API keys, exposed tokens, and fake credit card details
+    // E2E test with real data
+    // this.timeout(10000); // 10 seconds, since there are multiple API calls
```
jchen293, line 72: Can this comment be removed?

.gitignore
```diff
@@ -14,3 +14,5 @@ dist
 coverage
 
 build.tar.gz
+/test/cassettes/Referral*/creates-a-referral-user*
+/test/cassettes/Referral*/add-a-referral-user-credit-card*
```
Justintime50, line 18: As long as the tests are skipped, I don't see a need to include this. It frankly could frustrate the next person coming along trying to commit these (when we are able to redact info) and they can't find why these disappeared. Thoughts?

src/beta/resources/referral.js
```diff
@@ -0,0 +1,162 @@
...
+ * @param {string} creditCardExpirationYear - Credit card expiration year
+ * @param {string} creditCardCVC - Credit card CVC
+ * @returns {Promise<string>} - Stripe credit card ID
+ */
+async function sendCardDetailsToStripe(
+  api,
+  creditCardNumber,
+  creditCardExpirationMonth,
+  creditCardExpirationYear,
+  creditCardCVC,
```
Justintime50, line 55: We should unify the param names between Python/Node. I'd suggest that `creditCard` isn't needed on each of these since the name of the function already specifies this is for a credit card.

src/beta/resources/referral.js
```diff
@@ -0,0 +1,162 @@
...
+     * @param {string} referralUserId - The referral user's ID.
+     * @param {string} email - The new email address.
+     * @returns {Promise<boolean>} - Returns true if the referral was updated successfully, false otherwise.
+     */
+    static async updateEmail(referralUserId, email) {
+      try {
+        const newParams = { user: { email } };
+        await api.put(`${this._url}/${referralUserId}`, newParams);
+
+        return true;
```
Justintime50, line 122: We should be returning the message of the API, not a bool. This is how our delete methods across libs got all out of sync

src/beta/resources/referral.js
```diff
@@ -0,0 +1,162 @@
...
+     * @param {string} creditCardCVC - The credit card CVC.
+     * @param {string} primaryOrSecondary - Whether to add the card as 'primary' or 'secondary' payment method (defaults to 'primary')
+     * @returns {Promise<never>} - Response body (EasyPost payment method object)
+     */
+    static async addCreditCard(
+      referralApiKey,
+      creditCardNumber,
+      creditCardExpirationMonth,
+      creditCardExpirationYear,
+      creditCardCVC,
```
Justintime50, line 143: Same credit card comment as above. We'll want to unify the param names.

src/beta/resources/referral.js
Justintime50: To Jack's point, we should be wrapping these individual functions here in try/catch and returning helpful errors to the user in the off chance something happens outside our control (Stripe is down/times out, the keys of the request/response change, etc)

Justintime50 (changes requested):

src/beta/resources/referral.js
Justintime50: After discussing, I'm ok with returning `true` here as long as we try/catch in each lib so that we aren't blindly returning true (true will only get returned if no error).

src/beta/resources/referral.js
```diff
@@ -0,0 +1,162 @@
...
+ * @returns {Promise<string>} - Stripe credit card ID
+ */
+async function sendCardDetailsToStripe(
+  api,
+  creditCardNumber,
+  creditCardExpirationMonth,
+  creditCardExpirationYear,
+  creditCardCVC,
+) {
+  const stripeKey = await getEasyPostStripeKey(api);
```
Justintime50, line 57: Let's have this bubble up to the `addCreditCard` function so that that function can be the consolidator and each of these individual 3 steps/functions can only do one thing each (separation of concerns) instead of this function having the side affect of making two API calls and doing something that isn't in the function name.

Justintime50 (dismissed):

.gitignore
nwithan8: Sounds good

src/beta/resources/referral.js
```diff
@@ -0,0 +1,153 @@
...
+  email: T.string.isRequired,
+  phone: T.string.isRequired,
+});
+
+const stripeCreditCardUrl = 'https://api.stripe.com/v1/tokens';
+
+/**
+ * Get an instance of the API client using the referral user's API key.
+ * @param {API} api - The API client to copy.
+ * @param {string }referralApiKey - The referral user's API key.
```
jchen293, line 21:
```suggestion
 * @param {string} referralApiKey - The referral user's API key.
```

src/beta/resources/referral.js
```diff
@@ -0,0 +1,153 @@
...
+ * @returns {string} - The Stripe API key.
+ */
+async function getEasyPostStripeKey(api) {
+  const response = await api.get('partners/stripe_public_key');
+
+  return response.body.public_key;
+}
+
+/**
+ * Send the credit card details to Stripe to get a Stripe credit card ID.
```
jchen293, line 42:
```suggestion
 * Send the credit card details to Stripe to get a Stripe credit card token.
```

It should be Stripe credit card token, according to their doc: https://stripe.com/docs/api/tokens/create_card

src/beta/resources/referral.js
```diff
@@ -0,0 +1,153 @@
...
+ * @param {string} stripeKey - The Stripe API key.
+ * @param {string} number - Credit card number
+ * @param {string} expirationMonth - Credit card expiration month
+ * @param {string} expirationYear - Credit card expiration year
+ * @param {string} cvc - Credit card CVC
+ * @returns {Promise<string>} - Stripe credit card ID
```
jchen293, line 48: A missing period in the end, please also add period in other functions' docstring.
```suggestion
 * @param {string} number - Credit card number.
 * @param {string} expirationMonth - Credit card expiration month.
 * @param {string} expirationYear - Credit card expiration year.
 * @param {string} cvc - Credit card CVC.
 * @returns {Promise<string>} - Stripe credit card ID.
```

src/beta/resources/referral.js
```diff
@@ -0,0 +1,153 @@
...
+  } catch (error) {
+    if (error.response && error.response.body) {
+      throw new RequestError(error.response.body, stripeCreditCardUrl);
+    }
+    throw error;
+  }
+}
+
+/**
+ * Send the Stripe credit card ID to EasyPost to add the card to the user's account.
```
jchen293, line 78:
```suggestion
 * Send the Stripe credit card token to EasyPost to add the card to the user's account.
```

jchen293 (changes requested):
Code looks good, just some minor changes

src/beta/resources/referral.js
```diff
@@ -0,0 +1,153 @@
...
+  }
+}
+
+/**
+ * Send the Stripe credit card token to EasyPost to add the card to the user's account.
+ * @param {API} api - The API client to use.
+ * @param {string} referralApiKey - The referral user's production API key.
+ * @param {string} stripeCreditCardId - Stripe credit card ID
+ * @param {string} priority - Whether to add the card as the 'primary' or 'secondary' card
+ * @returns {Object} - Response body (EasyPost payment method object)
```
jchen293, line 83:
```suggestion
 * @param {string} stripeCreditCardId - Stripe credit card ID.
 * @param {string} priority - Whether to add the card as the 'primary' or 'secondary' card.
 * @returns {Object} - Response body (EasyPost payment method object).
```

src/beta/resources/referral.js
```diff
@@ -0,0 +1,153 @@
...
+
+    /**
+     * Add a credit card to the referral's account
+     * @param {string} referralApiKey - The referral user's production API key.
+     * @param {string} number - The credit card number.
+     * @param {string} expirationMonth - The credit card expiration month.
+     * @param {string} expirationYear - The credit card expiration year.
+     * @param {string} cvc - The credit card CVC.
+     * @param {string} primaryOrSecondary - Whether to add the card as 'primary' or 'secondary' payment method (defaults to 'primary')
+     * @returns {Promise<never>} - Response body (EasyPost payment method object)
```
jchen293, line 124:
```suggestion
     * @param {string} primaryOrSecondary - Whether to add the card as 'primary' or 'secondary' payment method (defaults to 'primary').
     * @returns {Promise<never>} - Response body (EasyPost payment method object).
```

jchen293 (approved):

src/beta/resources/referral.js
```diff
@@ -0,0 +1,153 @@
...
+    },
+  });
+
+  try {
+    const response = await request;
+
+    return response.body.id;
+  } catch (error) {
+    if (error.response && error.response.body) {
+      throw new RequestError(error.response.body, stripeCreditCardUrl);
```
Justintime50, line 71: Nit: did we want to match the Python lib's approach where we return our own message since whatever stripe returns us is out of our control?

nwithan8: Sure

Justintime50 (approved):

src/beta/resources/referral.js
```diff
@@ -0,0 +1,153 @@
...
+}
+
+/**
+ * Send the credit card details to Stripe to get a Stripe credit card token.
+ * @param {string} stripeKey - The Stripe API key.
+ * @param {string} number - Credit card number.
+ * @param {string} expirationMonth - Credit card expiration month.
+ * @param {string} expirationYear - Credit card expiration year.
+ * @param {string} cvc - Credit card CVC.
+ * @returns {Promise<string>} - Stripe credit card ID.
```
jchen293, line 48:
```suggestion
 * @returns {Promise<string>} - Stripe credit card token.
```

src/beta/resources/referral.js
```diff
@@ -0,0 +1,153 @@
...
+ * Send the Stripe credit card token to EasyPost to add the card to the user's account.
+ * @param {API} api - The API client to use.
+ * @param {string} referralApiKey - The referral user's production API key.
+ * @param {string} stripeCreditCardId - Stripe credit card ID.
+ * @param {string} priority - Whether to add the card as the 'primary' or 'secondary' card.
+ * @returns {Object} - Response body (EasyPost payment method object).
+ */
+async function sendCardDetailsToEasyPost(api, referralApiKey, stripeCreditCardId, priority) {
+  const _api = getReferralApi(api, referralApiKey);
+  const params = { credit_card: { stripe_object_id: stripeCreditCardId, priority } };
```
jchen293, line 87:
```suggestion
async function sendCardDetailsToEasyPost(api, referralApiKey, stripeCreditCardToken, priority) {
  const _api = getReferralApi(api, referralApiKey);
  const params = { credit_card: { stripe_object_id: stripeCreditCardToken, priority } };
```

jchen293 (approved):

jchen293 (approved):

Justintime50 (approved):

PR #244, opened 2022-05-02
Fix error status parsing

nwithan8:
# Description

When processing an error message from the API, oftentimes the Node library was not parsing the proper attributes, resulting in a printed error message like "Status code undefined returned by the API", which obviously isn't helpful.
This PR fixes the logic parsing the error message and code from the API response into a local `RequestError` object, including handling the two possible locations an error code could be stored, depending on the type of response (standard HTTP error or an EasyPost Error object)

- Fix error status/code not processing correctly on API error

# Testing

E2E tested, all unit tests pass - specifically unit test for checking RequestError object

# Pull Request Type


Justintime50 (approved):
This seems ok as it's simply adding more potential error locations to choose from and keeps the current behavior as fallbacks. Thanks for splitting this up, made it much easier to read through and think on.

jchen293 (dismissed):

Justintime50 (approved):

PR #267, opened 2022-07-06
combine credit card and bank account functions into a generic one.

jchen293:
# Description

Support for the bank account in billing options
Move PaymentMethod and CreditCard functions to Billing class
Update unit tests

# Testing

E2E test `retrievePaymentMethods` and `deletePaymentMethod`, but tests are skipped due to lack of a real payment in tests.

# Pull Request Type


nwithan8 (dismissed):

src/resources/billing.js
```diff
@@ -0,0 +1,100 @@
...
+  object: T.string,
+  name: T.string,
+  last4: T.string,
+  expiration_month: T.string,
+  expiration_year: T.string,
+  brand: T.string,
+  country: T.string,
+  verified: T.bool,
+  bank_account: T.string,
+};
```
Justintime50, line 15: Issue: These should really be props of a `PaymentMethod` because there is no "billing" object

src/resources/billing.js
```diff
@@ -0,0 +1,100 @@
...
+     * delete a payment method
+     * @returns {Promise|Promise<never>}
+     */
+    static async deletePaymentMethod(primaryOrSecondary) {
+      const paymentInfo = await this.getPaymentInfo(primaryOrSecondary.toLowerCase());
+      const endpoint = paymentInfo[0];
+      const paymentMethodID = paymentInfo[1];
+
+      const res = await api.del(`${endpoint}/${paymentMethodID}`);
+      return res.body;
```
Justintime50, line 53: Issue: This returns nothing, I believe we determined we were going to return `true` if successful?

src/resources/billing.js
```diff
@@ -0,0 +1,100 @@
...
+    static _name = 'Billing';
+
+    static _url = 'billing';
+
+    static key = 'billing';
+
+    static propTypes = propTypes;
+
+    /**
+     * fund your EasyPost wallet by charging your primary or secondary payment method on file.
```
Justintime50, line 28: Issue: These docstrings (all of them in this PR) should start with capitols.

src/resources/billing.js
```diff
@@ -0,0 +1,100 @@
...
+      const res = await api.get('payment_methods');
+
+      if (res.body.id == null) {
+        throw new Error('Billing has not been setup for this user. Please add a payment method.');
+      }
+      return res.body;
+    }
+
+    /**
+     * Get payment info (type of the payment method and ID of the payment method)
```
Justintime50, line 70: Since we can't yet support internal functions, please add to the docstring that this function is intended for internal use only to help us identify it in the future and dissuade users from using it.

src/resources/billing.js
```diff
@@ -0,0 +1,100 @@
...
+        secondary: 'secondary_payment_method',
+      };
+
+      const paymentMethodToUse = paymentMethodMap[primaryOrSecondary];
+      let paymentMethodID;
+      let endpoint;
+      const errorString = 'The chosen payment method is not valid. Please try again.';
+
+      if (paymentMethodToUse !== undefined && paymentMethods[paymentMethodToUse].id !== undefined) {
+        paymentMethodID = paymentMethods[paymentMethodToUse].id;
```
Justintime50, line 86: Suggestion: you could move this right above the if statement and reuse it in the if statement line since it's reused there too.

Justintime50 (changes requested):

src/resources/billing.js
nwithan8: Atlanta
Boston
Columbus
Denver
Salt Lake City
Albany

... Oh, you meant capitals...

Justintime50: capitals* 😛

src/resources/billing.js
jchen293: do you mean the `paymentMethodID` variable in line 86? We need to check if there is an ID exists in before assigning the variable value.

jchen293: I ran the test again with and without a payment method, we should check if there is a payment method in the response body before assigning the value to `paymentMethodID` variable to avoid this potential error `Cannot read properties of null (reading 'id')` when there the chosen payment method does not exist.

Please see my latest change

src/resources/billing.js
jchen293: fixed

src/resources/billing.js
jchen293: added

src/resources/billing.js
jchen293: added

nwithan8 (approved):

src/easypost.js
```diff
@@ -9,7 +9,7 @@ import Batch, { propTypes as batchPropTypes } from './resources/batch';
 import Brand, { propTypes as brandPropTypes } from './resources/brand';
 import CarrierAccount, { propTypes as carrierAccountPropTypes } from './resources/carrier_account';
 import CarrierType, { propTypes as carrierTypePropTypes } from './resources/carrier_type';
-import CreditCard, { propTypes as creditCardPropTypes } from './resources/credit_card';
+import Billing, { propTypes as BillingPropTypes } from './resources/billing';
```
Justintime50, line 12: Issue: This is not correct as the actual propTypes got moved to `PaymentMethod`, please correct this.

src/resources/payment_method.js
```diff
@@ -1,57 +1,20 @@
 import T from 'proptypes';
-import base from './base';
 
 export const propTypes = {
   id: T.string,
+  disabled_at: T.string,
```
Justintime50, line 5: Nit: object should come before disabled_at

Justintime50 (changes requested):

src/easypost.js
jchen293: `paymentMethodPropTypes` is down below in line 19

src/resources/payment_method.js
jchen293: fixed

nwithan8 (approved):

Justintime50 (approved):

PR #284, opened 2022-07-26
feat: adds validateWebhook function

Justintime50:
# Description

Adds a new `validateWebhook` function which takes in an `event_body` and `headers` from an EasyPost webhook and your locally stored `webhook_secret` and ensures the webhook originated from EasyPost. Will raise an error if the secrets don't match and produce your webhook if they do.

# Testing

Adds 3 unit tests to ensure all branches of this function work as intended.

# Pull Request Type

### DIFF
--- a/test/resources/webhook.test.js
+++ b/test/resources/webhook.test.js
@@ -1,8 +1,9 @@
 /* eslint-disable func-names */
 import { expect } from 'chai';
-import * as setupPolly from '../helpers/setup_polly';
+
 import EasyPost from '../../src/easypost';
 import Fixture from '../helpers/fixture';
+import * as setupPolly from '../helpers/setup_polly';
 
 describe('Webhook Resource', function () {
   setupPolly.startPolly();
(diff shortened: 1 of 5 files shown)


types/Webhook/Webhook.d.ts
```diff
@@ -71,4 +71,18 @@ export declare class Webhook implements IWebhook {
...
+   * Validate a webhook by comparing the HMAC signature header sent from EasyPost to your shared secret.
+   * If the signatures do not match, an error will be raised signifying the webhook either did not originate
+   * from EasyPost or the secrets do not match. If the signatures do match, the `event_body` will be returned
+   * as JSON.
+   *
+   * @param eventBody The event body of the webhook sent from EasyPost.
+   * @param headers The headers of the webhook sent from EasyPost.
+   * @param webhookSecret The local webhook secret that should match what is stored with EasyPost for this webhook.
+   *
+   * @see https://www.easypost.com/docs/api/node#delete-a-webhook
```
jchen293, line 85: The URL here is incorrect, it should be replaced it with `TODO`

jchen293 (approved):
I also E2E tested this with live webhooks, works as expected.

types/Webhook/Webhook.d.ts
Justintime50: Good catch. I've removed it.

test/resources/webhook.test.js
```diff
@@ -82,4 +83,43 @@ describe('Webhook Resource', function () {
 
     expect(deletedWebhook).to.be.an.instanceOf(this.easypost.Webhook);
   });
+
+  it('validates a webhook secret', function () {
+    const webhookSecret = 'sécret';
+    const expectedHmacSignature =
+      'hmac-sha256-hex=e93977c8ccb20363d51a62b3fe1fc402b7829be1152da9e88cf9e8d07115a46b';
```
nwithan8, line 90: Not: if this never changes, should we store it as a fixture?

nwithan8 (approved):

test/resources/webhook.test.js
Justintime50: Typically fixtures are for input data, not assertions/expectations. There would also not be a very concise way of saying "this signature belongs to this data over here"

jchen293 (approved):

PR #286, opened 2022-07-29
chore: looser params for verify and verify_strict

Justintime50:
# Description

Allows for looser values to the `verify` and `verify_strict` params when creating an address (can accept strings or bools outside of an array)

# Testing

- Updates old unit tests to use a bool
- Adds a new unit test to ensure arrays still work

# Pull Request Type

### DIFF
--- a/src/resources/address.js
+++ b/src/resources/address.js
@@ -1,4 +1,5 @@
 import T from 'proptypes';
+
 import base from './base';
 import { baseAddress as baseAddressPropTypes } from './base_address';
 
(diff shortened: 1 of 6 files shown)


src/resources/address.js
```diff
@@ -8,8 +9,8 @@ export const propTypes = Object.assign({}, baseAddressPropTypes, {
   carrier_facility: T.string,
   federal_tax_id: T.string,
   state_tax_id: T.string,
-  verify: T.array,
-  verify_strict: T.array,
+  verify: T.oneOfType([T.bool, T.string, T.array]),
```
nwithan8, line 12: Thought: Same question as on Python, shouldn't we be guiding users to only using booleans, rather than allowing them to continue doing it "the bad way"?

nwithan8 (changes requested):

src/resources/address.js
Justintime50: See my reply on the Python PR. Using arrays isn't "bad", there's just a more concise way via the Boolean.

nwithan8 (approved):

PR #293, opened 2022-08-19
Move EndShipper to general availability

nwithan8:
# Description

- Moved EndShipper from beta to main namespace
  - We can't have two classes with the same name/path (EndShipper was always `easypost.EndShipper`, not `easypost.beta.EndShipper`), so the beta class has been deleted
- Update unit tests

# Testing

- Unit tests updated and passing

# Pull Request Type

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@easypost/api",
-  "version": "5.5.0",
+  "version": "5.3.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@easypost/api",
-      "version": "5.5.0",
+      "version": "5.3.0",
       "license": "MIT",
       "dependencies": {
         "core-js": "3.19.3",
@@ -53,12 +53,12 @@
         "eslint-plugin-jest": "~22.21.0",
         "eslint-plugin-jsx-a11y": "^6.5.1",
         "eslint-plugin-react": "^7.29.3",
-        "mocha": "^10.0.0",
+        "mocha": "^8.4.0",
         "nyc": "^15.1.0",
         "prettier": "^2.5.1",
         "typescript": "4.4.4",
         "vows": "^0.8.3",
-        "webpack": "^5.74.0",
+        "webpack": "^4.44.1",
         "webpack-cli": "^3.3.12",
         "webpack-node-externals": "^1.7.2"
       },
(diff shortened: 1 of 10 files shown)


src/easypost.js
```diff
@@ -12,6 +12,7 @@ import CarrierType, { propTypes as carrierTypePropTypes } from './resources/carr
 import Billing, { propTypes as BillingPropTypes } from './resources/billing';
 import CustomsInfo, { propTypes as customsInfoPropTypes } from './resources/customs_info';
 import CustomsItem, { propTypes as customsItemPropTypes } from './resources/customs_item';
+import EndShipper from './resources/end_shipper';
```
Justintime50, line 15: Issue: We need the endShipperPropTypes here and then include them in the `PROP_TYPES` constant below.

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@easypost/api",
-  "version": "5.3.0",
+  "version": "5.5.0",
```
Justintime50, line 3: 🤷‍♂️  not a necessary change. The lockfile basically is a snapshot of what/when dependencies were installed/updated. The version of the lib here doesn't affect anything and serves as a moment in time to know when last all the things got bumped (2 versions ago). It's the version in the `package.json` file that is important but isn't going to get bumped just yet.

src/easypost.js
Justintime50, line 15: Still need those prop types imported here and then used in the `PROP_TYPES` constant below.

Justintime50 (changes requested):

package-lock.json
nwithan8: Is this not part of our .gitignore, and should it be?

src/easypost.js
nwithan8: Resolved.

package-lock.json
Justintime50: No, you definitely want lock files checked into the repo. This is what ensures that what you and I install on separate machines is the exact same. Because in the `package.json` file we do loose version pinning so that the next time we run `npm update` it updates the dependencies based on how loose our pinning is and then updates the lockfile to reflect that. But in-between dependency bumps, it will keep a consistent set of dependencies for everyone, everytime they do `npm install` - which is ideal. Python and other languages have similar constructs but we don't use them (with the exception of Go which is the `go.sum` file and in PHP with the `composer.lock` file).

Justintime50 (approved):

package-lock.json
nwithan8: So, do I need to roll this file back or not?

nwithan8: I take your approval as a no.

Justintime50: The change doesn't hurt, doesn't help - would suggest not making it in the future as it doesn't provide value but won't block on it 🙂

jchen293 (approved):

PR #308, opened 2022-10-21
chore: remove browser check when building headers

Justintime50:
# Description

Removes the browser check when building headers since this library out of the box cannot run in a browser.

# Testing

We tested using this library in the browser which is not possible, leaving me with high confidence that this is legacy code that can be removed easily.

# Pull Request Type

### DIFF
--- a/src/easypost.js
+++ b/src/easypost.js
@@ -148,21 +148,6 @@ export default class API {
   }
 
   /**
-   * Returns true if this code is running in a browser.
-   * @returns bool
-   */
-  static runningInBrowser() {
-    // TODO: This function may not be necessary any longer as we may block browser based requests, needs investigation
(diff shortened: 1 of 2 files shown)


src/easypost.js
```diff
@@ -173,12 +158,6 @@ export default class API {
       ...additionalHeaders,
     };
 
-    if (API.runningInBrowser()) {
```
Justintime50, line 176: # Reviewer Note

I accidentally introduced this function but had it flipped (it would remove headers for Node requests and keep them for browser which was backwards); however, this didn't get released yet so it was caught in time.

jchen293 (approved):

nwithan8 (approved):

PR #330, opened 2023-01-17
feat: Add functions to retrieve event payload(s)

nwithan8:
# Description

- New static functions to retrieve payload(s) for an event
- New Payload model and TypeScript definition
- New unit tests, cassettes to test payload functions
- New recordings for other event unit tests (oops)
- Improvements to docstrings

Since a Payload object does not have any instance functions (and as a small POC of the upcoming redesign), the Payload class does not extend the base class. As a result, there are a handful of functions (e.g. construction from JSON) that had to be ported out of the base class and re-implemented in the Payload class. It was either that, or extend the base class and override the various instance functions (`save`, `delete`, `all`) to not be implemented. In my opinion, this is a cleaner, safer approach.

The JSON of the response to retrieving all payloads technically returns an object with a single `payloads` key, like so:

```json
{
  "payloads": []
}
```

If this object were deserialized exactly, it would actually create a PayloadsCollection object (similar to what we do with Shipment Collection, Tracker Collection, etc. in type-strict languages). The actual Payload objects would be accessed by `payloads = payloadCollection.payloads`. This makes sense when the results are paginated, and the collection has more than one key (`has_more`). But since Payloads are not paginated, this doesn't make much sense. Instead, the retrieve all function remaps the data to simply return an array of Payload objects, which is easier to work with for the end-user and more accurately matches expectations.

# Testing

- New unit tests, cassettes

# Pull Request Type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## Next release:
 
 - Add `all` for retrieve all pickups in Pickup class.
+- Add `retrievePayload` and `retrieveAllPayloads` functions to retrieve payloads for an Event.
+  - These are static functions on the Event class, so you can call them like `easypost.Event.retrieveAllPayloads(eventId)`.
 
 ## v5.9.0 (2023-01-11)
 
--- a/src/resources/event.js
+++ b/src/resources/event.js
@@ -1,44 +1,75 @@
 import T from 'proptypes';
 import base from './base';
+import {Payload} from "./payload";
 
 export const propTypes = {
-  id: T.string,
-  object: T.string,
-  mode: T.string,
-  description: T.string,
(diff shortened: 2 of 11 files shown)


Justintime50 (changes requested):
I don't see that the payload prop types were ever added to the global constant `PROP_TYPES`. Please do that here.

nwithan8:
> I don't see that the payload prop types were ever added to the global constant `PROP_TYPES`. Please do that here.

Not needed, as the base class will never be accessing the properties or Payload class to deserialize (deserialization is done one-off inside the Payload class, not by the base class).

Justintime50 (approved):
🤨 trusting

jchen293 (approved):

PR #345, opened 2023-02-03
Thread Safe Rewrite - Data Creation Flow

Justintime50:
# Description

We set out to make all our libraries thread-safe. Come to find out Node is already thread-safe! This PR makes the additional changes that have accompanied the other library rewrites, namely the data creation flow. This means that in Node, instead of creating a local object and sending it to the API, you will instead send the params directly to the new `create` and `update` functions which make the API call.

This PR also fixes various other related bugs such as the double wrapping of results when retrieving a list (eg: `all`), and nested objects not deserializing properly.

- Renamed default export `API` to `EasyPostClient`
- Overhauled the data flow of the library
  - Instead of creating a local object and then calling `.save()` on it to create/update it at the API level, you will pass in the same data as before but to one of either `create()` or `update()` functions on the service in question
  - Services no longer implement the `delete()` function by default but instead implement it on a service-by-service basis
- Moved library structure around
  - Introduced `/models` that contain the EasyPost objects
  - Renamed `/resources` to `/services` to better reflect that a service called against an EasyPostClient differs from an API resource, now known as the `models`
  - Each service file now has `_service` appended
  - Added `/utils` which contains both `internal_util.js` which is not intended for user consumption and `util.js` which contains public utilities
  - Moved `getLowestSmartRate` and `validateWebook` function to `util.js` as neither of them require the client object to function
- References of `Referral` were changed to `ReferralCustomer` to match the API
- Corrects references of `smartrate` to `SmartRate` and `smartRate` to match the API
- Fixes the `verifyAddress` function to actually verify an address
- Fixes a bug that could double wrap or unintentionally wrap the results of an `/all` API call with the name of the object in question
- Completely overhauled deserialization process, nested objects (eg: rates of shipments, etc) now properly convert to their appropriate EasyPostObject

# Testing

- Adjusts all test syntax to match
- Re-records all cassettes
- Fixes skipped tests so they can run
- For whatever reason, certain changes introduced here necessitate skipping two scanform tests due to cassette mismatching... can revisit in the future

# Pull Request Type


src/models/pickup.js
```diff
@@ -0,0 +1,16 @@
...
+import EasyPostObject from './easypost_object';
+
+export default class Pickup extends EasyPostObject {
+  /**
+   * Get the lowest rate of a pickup.
+   * @param {string} carriers
+   * @param {string} services
+   * @returns {Object}
+   */
+  lowestRate(carriers, services) {
```
jchen293, line 11: nit: I would add an empty array default parameter to both carriers and services. This applies to other instance `lowestRate` functions

src/services/base_service.js
```diff
@@ -0,0 +1,232 @@
...
+
+      return convertedObject;
+    }
+
+    /**
+     * Creates a new NotImplementedError.
+     * @param {string} fnName
+     * @returns {Promise<never>}
+     */
+    static notImplemented(fnName) {
```
jchen293, line 158: nit: `fnName` is not a good name here

src/services/billing_service.js
```diff
@@ -87,4 +79,28 @@ export default (api) =>
...
+    /**
+     * create not implemented
+     * @returns {Promise<never>}
+     */
+    static create() {
+      return this.notImplemented('create');
+    }
+
+    /**
+     * create not implemented
```
jchen293, line 92:
```suggestion
     * retrieve not implemented
```

src/services/billing_service.js
```diff
@@ -87,4 +79,28 @@ export default (api) =>
...
+     * @returns {Promise<never>}
+     */
+    static retrieve() {
+      return this.notImplemented('retrieve');
+    }
+
+    /**
+     * create not implemented
```
jchen293, line 100:
```suggestion
     * all not implemented
```

src/services/rate_service.js
```diff
@@ -0,0 +1,26 @@
...
+    /**
+     * all not implemented
+     * @returns {Promise<never>}
+     */
+    static all() {
+      return this.notImplemented('all');
+    }
+
+    /**
+     * save not implemented
```
jchen293, line 20:
```suggestion
     * create not implemented
```

src/services/shipment_service.js
```diff
@@ -0,0 +1,199 @@
...
+        return Promise.reject(e);
+      }
+    }
+
+    /**
+     * Buy a shipment.
+     * @param {string} id
+     * @param {object} rate
+     * @param {number} insuranceAmount
+     * @param {boolean} withCarbonOffset
```
jchen293, line 38: missing docstring for `endShipperId`

src/services/shipment_service.js
```diff
@@ -0,0 +1,199 @@
...
+      } catch (e) {
+        return Promise.reject(e);
+      }
+    }
+
+    /**
+     * Get the lowest smartrate of this shipment.
+     * @param {string} id
+     * @param {number} delivery_days
+     * @param {string} delivery_accuracy
```
jchen293, line 186:
```suggestion
     * @param {number} deliveryDays
     * @param {string} deliveryAccuracy
```
These should be in camelCase

src/utils/util.js
```diff
@@ -0,0 +1,102 @@
+const crypto = require('crypto');
+
+module.exports = class Util {
+  /**
+   * Get the lowest smartrate of this shipment.
+   * @param {Object} smartrates
+   * @param {number} delivery_days
+   * @param {string} delivery_accuracy
```
jchen293, line 8:
```suggestion
   * @param {number} deliveryDays
   * @param {string} deliveryAccuracy
```
These should be in camelCase

src/easypost.js
```diff
@@ -51,63 +50,34 @@ export const METHODS = {
...
-  shipmentPropTypes,
-  trackerPropTypes,
-  userPropTypes,
-  webhookPropTypes,
-};
-
-export default class API {
+export default class EasyPostClient {
   constructor(key, options = {}) {
     const { useProxy, timeout, baseUrl, superagentMiddleware, requestMiddleware } = options;
```
jchen293, line 82: Question: Could we add `PropTypes` to validate these parameters?

jchen293 (changes requested):

src/models/pickup.js
Justintime50: Nice catch

src/services/base_service.js
Justintime50: Was here before, will make it verbose

src/easypost.js
Justintime50: There is a follow up task for this

src/models/pickup.js
jchen293: why did this get reverted?

src/models/order.js
```diff
@@ -0,0 +1,16 @@
...
+import EasyPostObject from './easypost_object';
+
+export default class Order extends EasyPostObject {
+  /**
+   * Get the lowest rate of an order.
+   * @param {string} carriers
+   * @param {string} services
+   * @returns {Object}
+   */
+  lowestRate(carriers, services) {
```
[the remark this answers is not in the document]
Justintime50: I rolled it back for all of them. It didn't like adding those.

src/models/pickup.js
Justintime50: I rolled it back for all of them. Build failed when these got changed

src/services/base_service.js
```diff
@@ -0,0 +1,232 @@
...
+    static notImplemented(functionName) {
+      return Promise.reject(new NotImplementedError(functionName, this._url));
+    }
+
+    /**
+     * Creates an EasyPost Object via the API.
+     * @param {*} params
+     * @returns {Base}
+     */
+    static async create(params) {
```
nwithan8, line 167: There's something to be said for not having any "create", "all" or "retrieve" functions in the base class (that you have to override to disable in subclasses), and instead re-implementing these functions directly in the appropriate service classes.

src/services/api_key_service.js
```diff
@@ -0,0 +1,8 @@
+import baseService from './base_service';
+
+export default (easypostClient) =>
+  class ApiKeyService extends baseService(easypostClient) {
```
nwithan8, line 4: "create", "all" and "retrieve" have been inherited from the base class and not overridden. That would mean there is a function to "create an API key", which is incorrect (see my comment on the base service re: inherited functions)

src/services/address_service.js
```diff
@@ -0,0 +1,78 @@
...
+    static _url = 'addresses';
+
+    static key = 'address';
+
+    /**
+     * Create an address.
+     * @param {*} params
+     * @returns {Address}
+     */
+    static async create(params) {
```
nwithan8, line 16: When users access the address service via `myClient.Address`, they are accessing an instance of this `AddressService` class. All API functions in this class (and all service classes) should be instance-based, not static.

nwithan8 (changes requested):

src/services/base_service.js
```diff
@@ -0,0 +1,232 @@
...
+    static _name = null;
+
+    static key = null;
+
+    /**
+     * Converts a response and all its nested elements to its associated EasyPostObject.
+     * @param {*} response
+     * @returns {*}
+     */
+    static convertToEasyPostObject(response) {
```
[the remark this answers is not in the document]
Justintime50: Agreed, it's a bit magical - but it's a battle tested approach to the problem that "just works". Porting it from the other libs moved this along very quickly.

The consolation here is that nested object deserialization gets fixed with this method which is great.

src/services/api_key_service.js
Justintime50: This may be an outlier, most should state something isn't implemented. I'll add that here. 

Were there others you noticed?

src/services/base_service.js
Justintime50: True - I did remove `delete` from the base class. This was the least amount of work required since this is how it's been. For services that don't allow one of create/all/retrieve, we add the `notImplemented` function to those.

src/services/address_service.js
Justintime50: hmmm... indeed. This'll take a bit of work.

src/services/base_service.js
nwithan8: But is that how we want to continue doing it, having functions in here that are "not implemented"?

src/services/address_service.js
nwithan8: Per internal conversations, it seems like this is actually the correct way to do it because Node is magic.

src/services/base_service.js
nwithan8: For a user perspective, I think it's kind of ugly to have a function available, only for it to spit out "sorry, this actually doesn't do anything".

Justintime50: I'm not necessarily opposed to changing it, if we do, we should handle it separately in a new PR?

nwithan8: We can if you want, to avoid blocking this PR. I'll take it on.

src/services/tracker_service.js
```diff
@@ -0,0 +1,20 @@
...
+    static key = 'tracker';
+
+    /**
+     * Create trackers in bulk.
+     * @param {object} params
+     */
+    static async createList(params = {}) {
+      const newParams = { trackers: params };
+      const url = 'trackers/create_list';
+      await easypostClient.post(url, newParams);
```
jchen293, line 18:
```suggestion
    static async createList(params = {}) {
      const wrappedParams = { trackers: params };
      const url = 'trackers/create_list';
      const response = await easypostClient.post(url, wrappedParams);
      
      return response;
```
This endpoint is not documentated, I know you didn't write this but it might need a small tweak.

jchen293 (commented):
Very close

src/services/tracker_service.js
Justintime50: It doesn't return anything so like the `delete` functions, you'd just call it and can't do anything with the response since it's empty.

jchen293: Ah, yes. I forgot this endpoint doesn't return anything

jchen293 (approved):
lgtm, well done!

PR #348, opened 2023-02-14
Explicitly implement all available functions

nwithan8:
# Description

- Abstract base "create", "retrieve" and "all" functions, make protected with leading _ (per convention)
- Explicitly implement "create", "retrieve" and "all" functions as necessary in each class
- Update unit tests accordingly
- Delete not_implemented.js
- Fix some docstrings

# Testing

- All unit tests pass with existing cassettes
- Removed unit tests that checked "Not Implemented", which is no longer a thing

# Pull Request Type


src/services/address_service.js
```diff
@@ -57,6 +57,38 @@ export default (easypostClient) =>
...
+    static async all(params = {}) {
+      try {
+        const url = this._url;
+        const response = await easypostClient.get(url, params);
+
+        return this.convertToEasyPostObject(response.body);
+      } catch (e) {
+        return Promise.reject(e);
+      }
+    }
```
Justintime50, line 74: This should probably be defined in the base service and we should have an `all` function that implements the base `all` function. See PHP on how we can accomplish this. Much less code and we get the best of both worlds where we have a base function that gets used by individual services.

nwithan8: That goes counter to the point of this PR?

Justintime50: The purpose of the initiative was that each service had an explicit set of functions available to it without inheriting or needing to not implement each.

The underlying guts don't change for these functions, set a URL, make a request, convert, and reject if an error occurs. Because of that, we should consolidate all that in the base service and each individual service can then implement those few functions and pass in the needed params. We do this with all our other libs.

This would fix the drop in coverage because we wouldn't need to test every individual promise reject line. Doing it this way reduces code duplication while still addressing your initial concern of not implementing certain functions on services.

nwithan8: But we'd have to un-implement the "all" function for some services that don't have it.

src/services/report_service.js
```diff
@@ -6,25 +6,38 @@ export default (easypostClient) =>
...
+      return this._create(url, params);
+    }
+
+    /**
+     * Retrieve a list of all reports associated with the API key.
+     * @param {string} type
+     * @param {object} params
+     * @returns {Report[]}
+     */
+    static async all(type, params = {}) {
```
Justintime50, line 23: This is an unintended breaking change. Let's just pull `type` out of `params` like it already was. All our other libs continue doing this.

src/services/scan_form_service.js
```diff
@@ -8,17 +8,35 @@ export default (easypostClient) =>
...
      * Create a ScanForm.
-     * @param {object} prams
+     * @param {object} params
      * @returns {ScanForm}
      */
     static async create(params) {
-      try {
-        // TODO: We should re-implement the check here that wraps up params in `shipments` if the user didn't
-        const response = await easypostClient.post(this._url, params);
+      // TODO: We should re-implement the check here that wraps up params in `shipments` if the user didn't
```
Justintime50, line 15: Do you mind tackling this simple TODO? If you go to the commit before mine, you'll find the previous implementation. It just needed a little tweak to properly wrap up the params. Makes sense to do this now since we are in here messing with the function implementation.

If not, I can do it after, but saves us all a review.

src/services/scan_form_service.js
```diff
@@ -8,17 +8,35 @@ export default (easypostClient) =>
...
+      const wrappedParams = {};
+      wrappedParams[this.key] = params;
+
+      return this._create(url, wrappedParams);
+    }
+
+    /**
+     * Retrieve a list of all scan forms associated with the API key.
+     * @param {object} params
+     * @returns {ScanForm[]}
```
Justintime50, line 27: Reminder that list of objects isn't correct throughout this PR since the results are paginated

Justintime50 (commented):
This is much better, thanks for doing it! Found a few items I'd like your input on

src/services/address_service.js
```diff
@@ -14,29 +14,24 @@ export default (easypostClient) =>
      * @returns {Address}
```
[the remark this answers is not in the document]
nwithan8: Yes, will do separately.

src/services/carrier_account_service.js
```diff
@@ -11,25 +11,21 @@ export default (easypostClient) =>
...
     static async create(params) {
-      try {
-        const carrierAccountType = params.type;
+      const carrierAccountType = params.type;
 
-        if (!carrierAccountType) {
-          throw new Error('CarrierAccount type is not set');
-        }
+      if (!carrierAccountType) {
+        throw new Error('CarrierAccount type is not set'); // this throw is caught by the catch block below, never returned to the user
```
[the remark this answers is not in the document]
nwithan8: A throw here would trigger the catch statement below, which would cause this function to return a Promis.reject

src/services/customs_info_service.js
```diff
@@ -9,10 +9,26 @@ export default (easypostClient) =>
     static key = 'customs_info';
 
     /**
-     * all not implemented.
-     * @returns {Promise<never>}
+     * Create a customs info record.
```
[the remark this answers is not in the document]
nwithan8: Will address in the docstring task

src/services/customs_item_service.js
```diff
@@ -9,10 +9,26 @@ export default (easypostClient) =>
     static key = 'customs_item';
 
     /**
-     * all not implemented.
-     * @returns {Promise<never>}
+     * Create a customs item.
+     * @param {*} params
```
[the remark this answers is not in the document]
nwithan8: Good catch, will address in later task

src/services/carrier_account_service.js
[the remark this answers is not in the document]
Justintime50: This was tested? We just need to make sure we surface this to the user so they know where they goofed.

src/services/batch_service.js
```diff
@@ -116,4 +130,23 @@ export default (easypostClient) =>
...
+    }
+
+    /**
+     * Retrieve a batch from the API.
+     * @param {string} id
+     * @returns {Batch}
+     */
+    static async retrieve(id) {
+      const url = `${this._url}/${id}`;
+      return this._retrieve(url);
```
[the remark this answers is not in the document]
nwithan8: Resolved

src/services/report_service.js
nwithan8: Looks like you added this during the rewrite last week? You're saying you want it removed?
image

src/services/end_shipper_service.js
```diff
@@ -14,14 +14,9 @@ export default (easypostClient) =>
...
-      try {
-        const wrappedParams = { address: params };
-        const response = await easypostClient.post(this._url, wrappedParams);
-
-        return this.convertToEasyPostObject(response.body);
-      } catch (e) {
-        return Promise.reject(e);
-      }
+      const wrappedParams = { address: params };
+      const url = `${this._url}`;
```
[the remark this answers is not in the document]
nwithan8: Resolved

src/services/carrier_account_service.js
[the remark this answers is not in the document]
nwithan8: Resolved

src/services/carrier_type_service.js
```diff
@@ -12,8 +12,10 @@ export default (easypostClient) =>
      * @returns {Array|Promise<never>}
      */
     static async all(query = {}) {
+      const url = `${this._url}`;
```
Justintime50, line 15: All occurances of ``${this._url}`` should just be `this._url` since there are no other strings as a part of it. Then, if we did that, not really a reason to assign it to a url variable but instead could just pass it in to the HTTP call directly

src/services/report_service.js
Justintime50: Huh. If I added it, it was a mistake. Can you verify I actually did this or was it like this before?

TLDR: Yes, we should have this just pull out of `params` since you can also pass in all the other bits like IDs and timestamps

src/services/scan_form_service.js
```diff
@@ -12,9 +12,14 @@ export default (easypostClient) =>
...
      */
     static async create(params) {
-      // TODO: We should re-implement the check here that wraps up params in `shipments` if the user didn't
       const url = `${this._url}`;
 
+      // wraps up params in `shipments` if the user didn't do it
+      if (!params.shipments) {
+        // eslint-disable-next-line no-param-reassign
+        params = { shipments: params };
+      }
```
Justintime50, line 21: This isn't quite what I had in mind, the implementation before would pull out individual shipment IDs and assign them each to an `id` key (needed for the API). Can we reimplement that?

Justintime50 (changes requested):

src/services/carrier_type_service.js
nwithan8: I agree with the first part about string interpolation. The second part I'll counter with readability and consistency.

src/services/report_service.js
nwithan8: Current version does it this way: https://github.com/EasyPost/easypost-node/blob/6d013bb1948b2162a14528a1d354b7ddce461631/src/resources/report.js#L51

Justintime50: Dang... I'd suggest to put it in params and call it out in the CHANGELOG. All our other libs operate this way, it's much more extensible, and it would be super weird to have a single function that doesn't follow the pattern the rest of the lib does.

We also talked about this internally and I believe the consensus was to put it in `params`.

src/services/scan_form_service.js
nwithan8: Resolved

src/services/report_service.js
nwithan8: Tale as old as time, C# does it one way, Java/PHP the other:

https://github.com/EasyPost/easypost-csharp/blob/aac9873379f710f8774e59f1d70f1141adeb626b/EasyPost/Services/ReportService.cs#L55

https://github.com/EasyPost/easypost-java/blob/351090eb9b3d2b836f36b683bed6bcc5d46edce4/src/main/java/com/easypost/service/ReportService.java#L85

https://github.com/EasyPost/easypost-php/blob/4818cc05c1fba77052cfebff7fa68143f7397412/lib/EasyPost/Service/ReportService.php#L38

nwithan8: (Yes, I did sing that.)

nwithan8: Anyway, thoughts? Introduce a breaking change, or keep it the way it is, since apparently cross-library consistency is out the window currently.

Justintime50: I believe I answered your question above

Justintime50 (approved):
Looks good! Love the extra 2% coverage

src/services/report_service.js
Justintime50: Lul, does this happen on Java/PHP too?

Justintime50: OH, delete `type` from params after crafting the URL

Justintime50 (approved):

src/services/batch_service.js
```diff
@@ -70,12 +86,13 @@ export default (easypostClient) =>
 
     /**
      * Create a scanform for a batch.
-     * @param {integer} id
+     * @param {number} id
```
jchen293, line 89:
```suggestion
     * @param {string} id
```

src/services/batch_service.js
```diff
@@ -103,17 +121,40 @@ export default (easypostClient) =>
 
     /**
      * Buy a batch.
-     * @param {integer} id
+     * @param {number} id
```
jchen293, line 124:
```suggestion
     * @param {string} id
```

src/services/pickup_service.js
```diff
@@ -8,6 +8,20 @@ export default (easypostClient) =>
 
     static key = 'pickup';
 
+    /**
+     * Create a pickup.
+     * @param {*} params
```
jchen293, line 13:
```suggestion
     * @param {object} params
```
nit: I am seeing `object` and `*` for the `params` parameter, we should stick to one of them.

src/services/scan_form_service.js
```diff
@@ -8,17 +8,49 @@ export default (easypostClient) =>
...
-        return Promise.reject(e);
+      // wraps up params in `shipments` if the user didn't do it
+      // turn a list of shipment objects into a map of shipment ids
+      if (params.shipments) {
+        // eslint-disable-next-line no-param-reassign
+        params.shipments = params.shipments.map((s) => {
+          if (typeof s === 'string') {
+            return { id: s };
+          }
+          return { id: s.id };
```
jchen293, line 25:
```suggestion
        params.shipments = params.shipments.map((shipment) => {
          if (typeof shipment === 'string') {
            return { id: shipment };
          }
          return { id: shipment.id };
```
nit: `s` is a bad naming for future travelers

src/services/user_service.js
```diff
@@ -84,15 +100,17 @@ export default (easypostClient) =>
 
     /**
      * Update the brand of a user.
-     * @param {integer} id
+     * @param {number} id
```
jchen293, line 103:
```suggestion
     * @param {string} id
```

jchen293 (commented):
some nitpicks, but we need to fix the docstring for the type of `id`.

nwithan8:
> some nitpicks, but we need to fix the docstring for the type of `id`.

Docstrings will be handled in a separate PR.

jchen293 (approved):

PR #350, opened 2023-02-17
make functions and fields private/protected if needed

jchen293:
# Description

make functions and fields private/protected

# Testing

Unit tests passed

# Pull Request Type

### DIFF
--- a/src/services/address_service.js
+++ b/src/services/address_service.js
@@ -2,9 +2,9 @@ import baseService from './base_service';
 
 export default (easypostClient) =>
   class AddressService extends baseService(easypostClient) {
-    static #_name = 'Address';
+    static #name = 'Address';
 
-    static #_url = 'addresses';
+    static #url = 'addresses';
 
     static #key = 'address';
 
@@ -14,7 +14,7 @@ export default (easypostClient) =>
      * @returns {Address}
      */
     static async create(params) {
-      const url = this.#_url;
+      const url = this.#url;
 
       const wrappedParams = {};
 
@@ -41,7 +41,7 @@ export default (easypostClient) =>
      * @returns {Address}
      */
     static async createAndVerify(params) {
-      const url = `${this.#_url}/create_and_verify`;
+      const url = `${this.#url}/create_and_verify`;
       const wrappedParams = { address: params };
 
       try {
@@ -59,7 +59,7 @@ export default (easypostClient) =>
      * @returns {Address[]}
      */
     static async all(params = {}) {
-      const url = this.#_url;
+      const url = this.#url;
 
       return this._all(url, params);
     }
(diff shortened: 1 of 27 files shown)


src/services/address_service.js
```diff
@@ -2,19 +2,19 @@ import baseService from './base_service';
 
 export default (easypostClient) =>
   class AddressService extends baseService(easypostClient) {
-    static _name = 'Address';
+    static #_name = 'Address';
```
Justintime50, line 5: We don't want these to be both pounded (#) and underscored. They should only be one or the other I'd think

Justintime50 (changes requested):
It looks like there's a bit of a mix of usage here. We can comb through this next week an make sure we catch all the items needed

src/services/address_service.js
jchen293: Oops, this is a typo here, but for the private fields it should be using `#`

nwithan8 (dismissed):
LGTM as long as you address what Justin brought up.

src/services/carrier_account_service.js
```diff
@@ -71,20 +71,20 @@ export default (easypostClient) =>
...
+    static _selectCarrierAccountCreationEndpoint(carrierAccountType) {
       if (Constants.CARRIER_ACCOUNTS_WITH_CUSTOM_WORKFLOWS.includes(carrierAccountType)) {
         return 'carrier_accounts/register';
       }
       return 'carrier_accounts';
     }
 
     /**
-     * Retrieve a list of all carrier accounts associated with the API key.
+     * Retrieve a list of all carrier accounts associated with the API #.
```
Justintime50, line 82:
```suggestion
     * Retrieve a list of all carrier accounts associated with the API key.
```

src/services/end_shipper_service.js
```diff
@@ -44,18 +44,18 @@ export default (easypostClient) =>
...
     static async retrieve(id) {
-      const url = `${this._url}/${id}`;
+      const url = `${this.#url}/${id}`;
 
       return this._retrieve(url);
     }
 
     /**
-     * Retrieve a list of all EndShippers associated with the API key.
+     * Retrieve a list of all EndShippers associated with the API #.
```
Justintime50, line 53:
```suggestion
     * Retrieve a list of all EndShippers associated with the API key.
```

src/services/shipment_service.js
```diff
@@ -209,12 +209,12 @@ export default (easypostClient) =>
     }
 
     /**
-     * Retrieve a list of all shipments associated with the API key.
+     * Retrieve a list of all shipments associated with the API #.
```
Justintime50, line 212:
```suggestion
     * Retrieve a list of all shipments associated with the API key.
```

Justintime50 (changes requested):
Nice, properties use pound, functions use underscore.

There were a couple instances where a docstring got clobbered with find and replace that needs fixing.

src/services/carrier_account_service.js
jchen293: Ooo, nice catch! This was unintended, search and replace gone wrong 😅

Justintime50 (approved):

PR #360, opened 2023-03-10
update style guide to use eslint:recommended

jchen293:
# Description

Removed `@easypost/eslint-config-easypost-base` as it's no longer maintained and is about to get deprecated, replaced the style guide with `eslint:recommended` and set up some basic config.

# Testing

# Pull Request Type

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -25,8 +25,7 @@
   },
   "parser": "@babel/eslint-parser",
   "parserOptions": {
-    "ecmaVersion": "latest",
-    "sourceType": "module"
+    "ecmaVersion": 13
   },
   "env": {
     "node": true,
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,6 @@
 {
   "extends": [
-    "@easypost/eslint-config-easypost-base",
+    "eslint:recommended",
     "prettier",
     "prettier/@typescript-eslint",
     "prettier/babel",
@@ -20,6 +20,16 @@
     "import/ignore": [".ejs$", ".yml$", ".jsx", "node_modules"]
   },
   "rules": {
-    "lines-between-class-members": "off"
+    "lines-between-class-members": "off",
+    "global-require": "off"
+  },
+  "parser": "@babel/eslint-parser",
+  "parserOptions": {
+    "ecmaVersion": "latest",
+    "sourceType": "module"
+  },
+  "env": {
+    "node": true,
+    "es6": true
   }
 }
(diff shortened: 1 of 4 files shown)


nwithan8 (approved):

test/.eslintrc
```diff
@@ -33,10 +33,16 @@
...
+    "jest/valid-expect": 0,
+    "global-require": "off"
   },
 
   "globals": {
     "expect": true
+  },
+  "parser": "@babel/eslint-parser",
+  "parserOptions": {
+    "ecmaVersion": "latest",
```
Justintime50, line 45: Issue: Please pin this `ecma` version. Without doing so, it will lead to lint errors when new rules or syntax becomes available.

.eslintrc
```diff
@@ -20,6 +20,16 @@
     "import/ignore": [".ejs$", ".yml$", ".jsx", "node_modules"]
   },
   "rules": {
-    "lines-between-class-members": "off"
+    "lines-between-class-members": "off",
+    "global-require": "off"
+  },
+  "parser": "@babel/eslint-parser",
+  "parserOptions": {
+    "ecmaVersion": "latest",
```
Justintime50, line 28: Issue: Please pin this `ecma` version. Without doing so, it will lead to lint errors when new rules or syntax becomes available.

.eslintrc
```diff
@@ -20,6 +20,16 @@
...
+    "sourceType": "module"
```
Justintime50, line 29: Question: I don't know that setting the sourceType here like this is appropriate since our library isn't setup as a module, though I know why you did it. Is there another option to fixing the `import` lint errors instead of saying this repo is a module which it's not?

test/.eslintrc
```diff
@@ -33,10 +33,16 @@
...
+    "sourceType": "module"
```
Justintime50, line 46: Question: I don't know that setting the sourceType here like this is appropriate since our library isn't setup as a module, though I know why you did it. Is there another option to fixing the `import` lint errors instead of saying this repo is a module which it's not?

Justintime50 (changes requested):

test/.eslintrc
jchen293: Pin to `13`

.eslintrc
jchen293: Actually, we don't need this anymore.

Justintime50 (approved):
This is a good start. I'm almost positive that a half dozen of our dev dependencies can also be removed since they were needed by the easypost lint rules but may not be needed anymore. We can explore that later.

nwithan8 (approved):

PR #363, opened 2023-03-24
Functionality to get the next page of a paginated collection

jchen293:
# Description

- Add `getNextPage` function that retrieves the next page of results for a paginated collection

# Testing

Add unit test to each paginated collection class

# Pull Request Type

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -71,4 +71,10 @@ export default class Constants {
    */
   static WEBHOOK_DOES_NOT_MATCH =
     'Webhook received did not originate from EasyPost or had a webhook secret mismatch.';
+
+  /**
+   * The template for the error message when a pagination list has reached its end
+   * @type {string}
+   */
+  static END_OF_PAGINATION = 'There are no more pages to retrieve.';
 }
--- a/src/services/address_service.js
+++ b/src/services/address_service.js
@@ -66,7 +66,7 @@ export default (easypostClient) =>
     }
 
     /**
-     * Retrieve the next page of specific collection of object
+     * Retrieve the next page of Address collection.
      * @param {Object} addresses An object containing a list of {@link Address addresses} and pagination information.
      * @param {Number} pageSize The number of records to return on each page
      * @returns {EasyPostObject|Promise<never>} The retrieved {@link EasyPostObject}-based class instance, or a `Promise` that rejects with an error.
--- a/src/services/address_service.js
+++ b/src/services/address_service.js
@@ -71,7 +71,7 @@ export default (easypostClient) =>
      * @param {Number} pageSize The number of records to return on each page
      * @returns {EasyPostObject|Promise<never>} The retrieved {@link EasyPostObject}-based class instance, or a `Promise` that rejects with an error.
      */
-    static async getNextPage(addresses, pageSize) {
+    static async getNextPage(addresses, pageSize = null) {
       const url = 'addresses';
       return this._getNextPage(url, addresses, pageSize);
     }
--- a/src/services/address_service.js
+++ b/src/services/address_service.js
@@ -73,8 +73,7 @@ export default (easypostClient) =>
      */
     static async getNextPage(addresses, pageSize) {
       const url = 'addresses';
-      const addressesArray = addresses.addresses;
-      return this._getNextPage(url, addressesArray, pageSize);
+      return this._getNextPage(url, addresses, pageSize);
     }
 
     /**
--- a/src/services/address_service.js
+++ b/src/services/address_service.js
@@ -66,6 +66,18 @@ export default (easypostClient) =>
     }
 
     /**
+     * Retrieve the next page of specific collection of object
+     * @param {Object} addresses An object containing a list of {@link Address addresses} and pagination information.
+     * @param {Number} pageSize The number of records to return on each page
+     * @returns {EasyPostObject|Promise<never>} The retrieved {@link EasyPostObject}-based class instance, or a `Promise` that rejects with an error.
+     */
+    static async getNextPage(addresses, pageSize) {
+      const url = 'addresses';
+      const addressesArray = addresses.addresses;
+      return this._getNextPage(url, addressesArray, pageSize);
+    }
+
+    /**
      * Retrieve an {@link Address address} by its ID.
      * See {@link https://www.easypost.com/docs/api/node#retrieve-an-address EasyPost API Documentation} for more information.
      * @param {string} id - The ID of the address to retrieve.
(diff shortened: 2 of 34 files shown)


nwithan8 (dismissed):

src/services/base_service.js
```diff
@@ -191,4 +192,26 @@ export default (easypostClient) =>
...
+    /**
+     * Retrieve the next page of specific collection of object
+     * @internal
+     * @param {string} url The URL to send the API request to.
+     * @param {Object} collectionArray The array of specific object.
+     * @param {Number} pageSize The number of records to return on each page
+     * @returns {EasyPostObject|Promise<never>} The retrieved {@link EasyPostObject}-based class instance, or a `Promise` that rejects with an error.
+     */
+    static async _getNextPage(url, collectionArray, pageSize) {
+      let params = {
```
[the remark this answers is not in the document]
nwithan8: Yep, override sounds good

nwithan8 (dismissed):
Please add a note that we need to add batches and EndShippers once the API supports them properly.

jchen293:
> 



> Please add a note that we need to add batches and EndShippers once the API supports them properly.

Do you mean a comment note or a `get_next_page` function that throws an error?

nwithan8:
> > 
> 
> > Please add a note that we need to add batches and EndShippers once the API supports them properly.
> 
> Do you mean a comment note or a `get_next_page` function that throws an error?

Just a comment for now, we got away from adding "not implemented" functions in Node during the rewrite.

nwithan8 (dismissed):

src/services/base_service.js
```diff
@@ -197,17 +197,14 @@ export default (easypostClient) =>
...
      * TODO: Implement this function in EndShippers and Batches once the API supports them properly.
      */
-    static async _getNextPage(url, collectionArray, pageSize) {
-      if (
-        collectionArray == undefined ||
-        collectionArray.length == 0 ||
-        !collectionArray.has_more
-      ) {
+    static async _getNextPage(url, collection, pageSize) {
+      const collectionArray = collection[url];
```
nwithan8, line 206: Please, no. This is just bound to break in the future.

nwithan8 (changes requested):

src/services/base_service.js
jchen293: Are we making the change in the API versioning?

nwithan8: Moving conversation internally.

nwithan8: Thanks for clearing that up for me internally. Won't block.

nwithan8 (dismissed):

src/constants.js
```diff
@@ -71,4 +71,10 @@ export default class Constants {
    */
   static WEBHOOK_DOES_NOT_MATCH =
     'Webhook received did not originate from EasyPost or had a webhook secret mismatch.';
+
+  /**
+   * The template for the error message when a pagination list has reached its end
+   * @type {string}
+   */
```
Justintime50, line 78: thought: I don't love that we decided to document every constant string. They should be self explanatory from the name and string. Lots of extra unnecessary comments imo

src/services/address_service.js
```diff
@@ -65,6 +65,17 @@ export default (easypostClient) =>
       return this._all(url, params);
     }
 
+    /**
+     * Retrieve the next page of specific collection of object
```
Justintime50, line 69: nit: I'd prefer we word all of these something like this:

```suggestion
     * Retrieve the next page of a collection of objects
```

Justintime50 (dismissed):

src/services/report_service.js
```diff
@@ -62,4 +74,10 @@ export default (easypostClient) =>
 
       return this._retrieve(url);
     }
+
+    static _getReportType(reportType) {
```
Justintime50, line 78: issue: This should have a docstring defining what it does. Regex can get messy so we should have a readable description of it

Also, I thought we had this functionality used elsewhere too for the normal function, do we need to redefine this here?

jchen293: The docstring is added. I don't see we have this helper function in the code.

Justintime50: Well done

src/services/report_service.js
```diff
@@ -43,12 +44,23 @@ export default (easypostClient) =>
      * Retrieve the next page of specific collection of object
      * @param {Object} reports An object containing a list of {@link Report reports} and pagination information.
      * @param {Number} pageSize The number of records to return on each page
-     * @param {string} type The type of report to be retrieved
      * @returns {EasyPostObject|Promise<never>} The retrieved {@link EasyPostObject}-based class instance, or a `Promise` that rejects with an error.
      */
-    static async getNextPage(reports, type, pageSize) {
+    static async getNextPage(reports, pageSize) {
+      const type = this._getReportType(reports.reports[0].object);
```
nwithan8, line 50: Might I recommend instead storing the `type` in the report collection object when it's first retrieved: https://github.com/EasyPost/easypost-csharp/blob/98b378530b8e9234678cfb7ebc7e9ce93f329720/EasyPost/Services/ReportService.cs#L70

And then you have access to it to reuse for subsequent API calls?
https://github.com/EasyPost/easypost-csharp/blob/98b378530b8e9234678cfb7ebc7e9ce93f329720/EasyPost/Services/ReportService.cs#LL82C236-L82C236

This would also need to extend to shipments and trackers, which have additional parameters that need to be stored and reused:
https://github.com/EasyPost/easypost-go/blob/3933664fef2385530e776dc7bda4e1d8a00214eb/shipment.go#L224
https://github.com/EasyPost/easypost-go/blob/3933664fef2385530e776dc7bda4e1d8a00214eb/tracker.go#L227

Pinging @Justintime50 on this too

nwithan8 (changes requested):

src/services/report_service.js
Justintime50: Either seem fine

jchen293: I mean either way works, storing `type` in the report collection will expose `type` which is something undocumented in the API Doc but I don't see any concern with this.

nwithan8:
Blocking this until the Shipment and Tracker parameters are sorted out, design-wise, on other CLs:
https://github.com/EasyPost/easypost-csharp/pull/424

src/services/report_service.js
jchen293: Fixed

nwithan8 (approved):

Justintime50 (approved):

PR #383, opened 2023-05-09
Moves `Utils` to EasyPostClient class

jchen293:
# Description

Moves `Utils` to EasyPostClient class so it can be exposed to the end users.
# Testing

# Pull Request Type

### DIFF
--- a/src/easypost.js
+++ b/src/easypost.js
@@ -32,6 +32,7 @@ import ShipmentService from './services/shipment_service';
 import TrackerService from './services/tracker_service';
 import UserService from './services/user_service';
 import WebhookService from './services/webhook_service';
+import Utils from './utils/util';
 
 const util = require('util');
 
@@ -108,6 +109,7 @@ export const SERVICES = {
   Tracker: TrackerService,
   User: UserService,
   Webhook: WebhookService,
+  Utils: Utils,
 };
 
 /**
--- a/src/models/order.js
+++ b/src/models/order.js
@@ -1,5 +1,5 @@
 /* eslint-disable babel/camelcase */
-import Utils from '../utils/util';
+import EasyPostClient from '../easypost';
 import EasyPostObject from './easypost_object';
 
 /**
@@ -29,6 +29,6 @@ export default class Order extends EasyPostObject {
   lowestRate(carriers, services) {
     const rates = this.rates || [];
 
-    return Utils.getLowestRate(rates, carriers, services);
+    return new EasyPostClient('fake_api_key').Utils.getLowestRate(rates, carriers, services);
   }
 }
--- a/src/models/pickup.js
+++ b/src/models/pickup.js
@@ -1,5 +1,5 @@
 /* eslint-disable babel/camelcase */
-import Utils from '../utils/util';
+import EasyPostClient from '../easypost';
 import EasyPostObject from './easypost_object';
 
 /**
@@ -32,6 +32,6 @@ export default class Pickup extends EasyPostObject {
   lowestRate(carriers, services) {
     const rates = this.pickup_rates || [];
 
-    return Utils.getLowestRate(rates, carriers, services);
+    return new EasyPostClient('fake_api_key').Utils.getLowestRate(rates, carriers, services);
   }
 }
--- a/src/models/shipment.js
+++ b/src/models/shipment.js
@@ -1,5 +1,5 @@
 /* eslint-disable babel/camelcase */
-import Utils from '../utils/util';
+import EasyPostClient from '../easypost';
 import EasyPostObject from './easypost_object';
 
 /**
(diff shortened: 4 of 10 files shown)


src/easypost.js
```diff
@@ -32,6 +32,7 @@ import ShipmentService from './services/shipment_service';
 import TrackerService from './services/tracker_service';
 import UserService from './services/user_service';
 import WebhookService from './services/webhook_service';
+import Utils from './utils/util';
```
Justintime50, line 35: suggestion: please move this to the right spot to be sorted

src/easypost.js
```diff
@@ -108,6 +109,7 @@ export const SERVICES = {
   Tracker: TrackerService,
   User: UserService,
   Webhook: WebhookService,
+  Utils: Utils,
```
Justintime50, line 112: suggestion: please move this to the right spot to be sorted

src/models/order.js
```diff
@@ -29,6 +29,6 @@ export default class Order extends EasyPostObject {
   lowestRate(carriers, services) {
     const rates = this.rates || [];
 
-    return Utils.getLowestRate(rates, carriers, services);
+    return new EasyPostClient('fake_api_key').Utils.getLowestRate(rates, carriers, services);
```
Justintime50, line 32: issue: we should not be constructing clients inside our own code. The client is for the user. Because we have direct access to the code before it gets compiled, we should still be able to reference the utils relatively. It's our users who cannot do this because they only get access to the single exposed client.

src/models/pickup.js
```diff
@@ -32,6 +32,6 @@ export default class Pickup extends EasyPostObject {
   lowestRate(carriers, services) {
     const rates = this.pickup_rates || [];
 
-    return Utils.getLowestRate(rates, carriers, services);
+    return new EasyPostClient('fake_api_key').Utils.getLowestRate(rates, carriers, services);
```
Justintime50, line 35: issue: we should not be constructing clients inside our own code. The client is for the user. Because we have direct access to the code before it gets compiled, we should still be able to reference the utils relatively. It's our users who cannot do this because they only get access to the single exposed client.

src/models/shipment.js
```diff
@@ -45,6 +45,6 @@ export default class Shipment extends EasyPostObject {
   lowestRate(carriers, services) {
     const rates = this.rates || [];
 
-    return Utils.getLowestRate(rates, carriers, services);
+    return new EasyPostClient('fake_api_key').Utils.getLowestRate(rates, carriers, services);
```
Justintime50, line 48: issue: we should not be constructing clients inside our own code. The client is for the user. Because we have direct access to the code before it gets compiled, we should still be able to reference the utils relatively. It's our users who cannot do this because they only get access to the single exposed client.

src/services/shipment_service.js
```diff
@@ -205,7 +205,11 @@ export default (easypostClient) =>
      */
     static async lowestSmartRate(id, deliveryDays, deliveryAccuracy) {
       const smartRates = await this.getSmartRates(id);
-      return Util.getLowestSmartRate(smartRates, deliveryDays, deliveryAccuracy.toLowerCase());
+      return new EasyPostClient('fake_api_key').Utils.getLowestSmartRate(
```
Justintime50, line 208: issue: we should not be constructing clients inside our own code. The client is for the user. Because we have direct access to the code before it gets compiled, we should still be able to reference the utils relatively. It's our users who cannot do this because they only get access to the single exposed client.

Justintime50 (changes requested):

nwithan8 (dismissed):

src/models/order.js
nwithan8: Well, that won't work then, because we don't want to create a new client like Justin said, but we also don't have access to an existing client inside a model.

Do we want to keep these instance functions? `myOrder.lowestRate()` versus `myClient.Utils.lowestRate(myOrder.rates)`?

jchen293: We should keep them instance function for the object, otherwise it would be a breaking change

src/utils/util.js
```diff
@@ -11,139 +12,140 @@ const util = require('util');
...
-
-    if (!validDeliveryAccuracyValues.has(lowercaseDeliveryAccuracy)) {
-      throw new InvalidParameterError({
-        message: `Invalid deliveryAccuracy value, must be one of: ${new Array(
-          ...validDeliveryAccuracyValues,
-        ).join(', ')}`,
-      });
-    }
+export default (easypostClient) =>
+  class Utils extends baseService(easypostClient) {
```
Justintime50, line 16: issue: this should not extend the `base_service` because we don't need anything from the base service here since none of these are making API calls.

Justintime50 (changes requested):

jchen293:
Closing this PR in favor of this [PR](https://github.com/EasyPost/easypost-node/pull/384) with the new branch name. All the feedback has been addressed in the new branch

PR #396, opened 2023-07-12
Adds request and response hooks to client

ralexmatthews:
# Description

Adds config options for an optional `requestHook` and `responseHook`. These functions provide various useful debugging values, outlined in the `.d.ts`

```ts
export interface IEasyPostRequest {
  method: 'get' | 'post' | 'put' | 'patch' | 'del';
  path: string;
  requestBody: any;
  requestHeaders: Record<string, any>;
  requestTimestamp: number;
  requestUUID: string;
}
export interface IEasyPostResponse extends IEasyPostRequest {
  httpStatus: number;
  responseBody: any;
  responseHeaders: Record<string, any>;
  responseTimestamp: number;
}
```

# Testing

I added a unit test to make sure when providing hooks, that the values are set to somewhat what we would expect them to be.

# Pull Request Type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v6.7 (2023-07-12)
+
+- Adds `requestHook` and `responseHook` config options
+
 ## v6.6.1 (2023-06-20)
 
 - Fixes the `pickup_rates` Typescript property of a Pickup
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@easypost/api",
   "description": "EasyPost Node Client Library",
-  "version": "6.6.1",
+  "version": "6.7",
   "author": "Easypost Engineering <oss@easypost.com>",
   "homepage": "https://easypost.com",
   "bin": {
@@ -38,6 +38,7 @@
     "regenerator-runtime": "~0.13.11",
     "source-map-support": "~0.5.21",
     "superagent": "~8.0.9",
+    "uuid": "^9.0.0",
     "yargs": "~17.7.2",
     "yargs-parser": "~21.1.1"
   },
--- a/src/easypost.js
+++ b/src/easypost.js
@@ -1,5 +1,6 @@
 import os from 'os';
 import superagent from 'superagent';
+import { v4 as uuid } from 'uuid';
 
 import pkg from '../package.json';
 import Constants from './constants';
@@ -120,7 +121,15 @@ export const SERVICES = {
  */
 export default class EasyPostClient {
   constructor(key, options = {}) {
-    const { useProxy, timeout, baseUrl, superagentMiddleware, requestMiddleware } = options;
+    const {
+      useProxy,
+      timeout,
+      baseUrl,
+      superagentMiddleware,
+      requestMiddleware,
+      requestHook,
+      responseHook,
+    } = options;
 
     if (!key && !useProxy) {
       throw new MissingParameterError({
@@ -133,6 +142,8 @@ export default class EasyPostClient {
     this.baseUrl = baseUrl || DEFAULT_BASE_URL;
     this.agent = superagent;
     this.requestMiddleware = requestMiddleware;
+    this.requestHook = requestHook;
+    this.responseHook = responseHook;
     this.Utils = new Utils();
 
     if (superagentMiddleware) {
@@ -221,16 +232,47 @@ export default class EasyPostClient {
       request.auth(this.key);
     }
 
-    if (params !== {} && params !== undefined) {
+    // it would be ideal if this "full url with params" could be gotten from superagent directly,
+    // but it doesn't seem to be possible
+    const url = new URL(urlPath);
+
+    if (params !== undefined) {
       if (method === METHODS.GET || method === METHODS.DELETE) {
         request.query(params);
+        Object.entries(params).forEach(([key, value]) => {
+          url.searchParams.append(key, value);
+        });
       } else {
         request.send(params);
       }
     }
 
+    const baseHooksValue = {
+      method,
+      path: url.toString(),
+      requestBody: request._data,
+      requestHeaders,
+      requestTimestamp: Date.now(),
+      requestUUID: uuid(),
+    };
+
+    if (this.requestHook) {
+      this.requestHook(baseHooksValue);
+    }
+
     try {
       const response = await request;
+
+      if (this.responseHook) {
+        this.responseHook({
+          ...baseHooksValue,
+          httpStatus: response.status,
+          responseBody: response.body,
+          responseHeaders: response.headers,
+          responseTimestamp: Date.now(),
+        });
+      }
+
       return response;
     } catch (error) {
       if (error.response && error.response.body) {
--- a/test/services/easypost.test.js
+++ b/test/services/easypost.test.js
@@ -1,7 +1,8 @@
 import { expect } from 'chai';
 
-import EasyPost from '../../src/easypost';
+import EasyPost, { METHODS } from '../../src/easypost';
 import MissingParameterError from '../../src/errors/general/missing_parameter_error';
+import Fixture from '../helpers/fixture';
 
 /* eslint-disable func-names */
 describe('EasyPost', function () {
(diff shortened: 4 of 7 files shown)


src/easypost.js
```diff
@@ -221,16 +232,47 @@ export default class EasyPostClient {
       request.auth(this.key);
     }
 
-    if (params !== {} && params !== undefined) {
```
ralexmatthews, line 224: Got rid of the `params !== {}` check because it is always true

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v6.7 (2023-07-12)
```
jchen293, line 3:
```suggestion
## Next release
```
We are not releasing this today, whenever we update the changelog, we typically just put `Next release` and then will have a separate PR to update the version and the date

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
...
+
+- Adds `requestHook` and `responseHook` config options
```
jchen293, line 5:
```suggestion
- Adds `requestHook` and `responseHook` config options in easypost client
```
For changelog entry, let's add a bit more information here

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@easypost/api",
   "description": "EasyPost Node Client Library",
-  "version": "6.6.1",
+  "version": "6.7.0",
```
jchen293, line 4: This need to be rollback, we will have a separate to update these version when we release it

test/services/easypost.test.js
```diff
@@ -11,4 +12,43 @@ describe('EasyPost', function () {
...
     );
   });
+
+  it('will log the appropriate values when a request and response hooks are provided', async function () {
+    let requestConfig;
+    const requestHook = (response) => (requestConfig = response);
+    let responseConfig;
+    const responseHook = (response) => (responseConfig = response);
+
+    const client = new EasyPost(process.env.EASYPOST_TEST_API_KEY, {
```
nwithan8, line 22: This unit test is throwing an error about an invalid API key. Is this the correct env var?

nwithan8 (changes requested):

src/easypost.js
```diff
@@ -221,16 +232,47 @@ export default class EasyPostClient {
...
       } else {
         request.send(params);
       }
     }
 
+    const baseHooksValue = {
+      method,
+      path: url.toString(),
+      requestBody: request._data,
+      requestHeaders,
```
jchen293, line 254:
```suggestion
      headers,
```

This should be called `headers` to follow our tech design

test/services/easypost.test.js
```diff
@@ -11,4 +12,43 @@ describe('EasyPost', function () {
...
+    expect(responseConfig.httpStatus).to.be.a('number');
+    expect(responseConfig.responseBody).to.be.an('object');
+    expect(responseConfig.responseBody.object).to.equal('Address');
+    expect(responseConfig.responseHeaders).to.be.an('object');
+    expect(responseConfig.responseHeaders['content-type']).to.contain('application/json');
+    expect(responseConfig.responseTimestamp).to.be.a('number');
+    expect(responseConfig.responseTimestamp).to.be.greaterThanOrEqual(
+      responseConfig.requestTimestamp,
+    );
+  });
```
jchen293, line 53: We need a test for unsubscribe request and response hooks. I believe we are missing unsubscribe functionality

jchen293 (changes requested):

src/easypost.js
ralexmatthews: Hmmm... do you think it might be a little bit confusing to have both `headers` and `responseHeaders` in the same object on the response? I renamed it to `requestHeaders` to help distinguish between the two.

test/services/easypost.test.js
ralexmatthews: Hmmm... yea I guess this is a little different from how it is done in the other diffs. Alrighty. I'll change this API to use `subscribe` and `unsubscribe` methods instead

src/easypost.js
jchen293: We should rename it to stay consistent across all CLs. The `responseHook` should not be getting `requestHeader`, you will need to make some changes in line 267 `...baseHooksValue`

nwithan8:
@jchen293 We've got a new CVE: https://github.com/EasyPost/easypost-node/actions/runs/5536807563/jobs/10104950886?pr=396

src/easypost.js
```diff
@@ -221,19 +266,62 @@ export default class EasyPostClient {
...
+      path: url.toString(),
+      requestBody: request._data,
+      headers: requestHeaders,
+      requestTimestamp: Date.now(),
+      requestUUID: uuid(),
+    };
+
+    if (this.requestHooks.length > 0) {
+      this.requestHooks.forEach((fn) => fn(baseHooksValue));
+    }
```
jchen293, line 295:
```suggestion
      this.requestHooks.forEach((fn) => fn(baseHooksValue));
```

Nit: If statement is not necessary here

test/services/easypost.test.js
```diff
@@ -1,14 +1,148 @@
...
     );
   });
+
+  it('will log the appropriate values when a request and response hooks are provided', async function () {
+    let requestConfig;
+    const requestHook = (response) => (requestConfig = response);
+    let responseConfig;
+    const responseHook = (response) => (responseConfig = response);
+
+    const client = new EasyPost(process.env.EASYPOST_TEST_API_KEY);
```
jchen293, line 25: Nit: instead of defining `client` every time in each unit test, you can define it globally. Example: https://github.com/EasyPost/easypost-node/blob/master/test/services/address.test.js#L14-L16

jchen293 (changes requested):

src/easypost.js
```diff
@@ -142,6 +145,48 @@ export default class EasyPostClient {
     this._attachServices(SERVICES);
   }
 
+  /**
+   * Add a request hook function.
+   * @param {(config: object) => void} hook
+   */
+  addRequestHook(hook) {
+    this.requestHooks = [...this.requestHooks, hook];
```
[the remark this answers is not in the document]
ralexmatthews: Ehhhh... I respectfully disagree. I tend to prefer a more functional, immutable approach over mutability.

test/services/easypost.test.js
ralexmatthews: 👍

src/easypost.js
ralexmatthews: 👍

src/easypost.js
```diff
@@ -221,19 +266,62 @@ export default class EasyPostClient {
...
       return response;
     } catch (error) {
       if (error.response && error.response.body) {
+        this.responseHooks.forEach((fn) =>
+          fn({
+            ...baseHooksValue,
+            httpStatus: error.response.status,
+            responseBody: error.response.body,
+            responseHeaders: error.response.headers,
+            responseTimestamp: Date.now(),
```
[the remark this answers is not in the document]
ralexmatthews: Made a function to create value

src/easypost.js
ralexmatthews: done

src/easypost.js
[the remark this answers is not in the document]
nwithan8: I agree with Alex. Plus, takes the grunt work away from us having to implement a working push-pop functionality to find and remove a specific hook. Let the user handle that.

src/easypost.js
nwithan8: How so? It defaults to `{}` and there are certainly HTTP call functions (e.g. retrieve) that don't have any query/body data and therefore no `params`.

nwithan8 (changes requested):

types/EasyPost.d.ts
```diff
@@ -16,6 +16,21 @@ import { User } from './User';
...
+  path: string;
+  requestBody: any;
+  headers: Record<string, any>;
+  requestTimestamp: number;
+  requestUUID: string;
+}
+export interface IEasyPostResponse extends Omit<IEasyPostRequest, 'headers'> {
+  httpStatus: number;
+  responseBody: any;
+  responseHeaders: Record<string, any>;
```
jchen293, line 30:
```suggestion
  headers: Record<string, any>;
```

jchen293 (commented):
one last comment on the typescript definition

README.md
```diff
@@ -161,12 +161,40 @@ const client = new EasyPostClient('my-key', {
...
 ```
 
+### Subscribing to internal requests and responses
+
+If you would need or want to view what requests are being made to the EasyPost API behind the scenes, you can provide functions to listen for the request and/or responses. This can be done with helper functions on the created client.
+
+```javascript
+const client = new EasyPostClient('my-key');
+
+const logOutgoingRequest = (request) => console.log("Outgoing:", requestData);
```
jchen293, line 177:
```suggestion
const logOutgoingRequest = (request) => console.log("Outgoing:", request);
```

jchen293 (dismissed):
lgtm, but there is a typo in the readme

types/EasyPost.d.ts
ralexmatthews: 👍

jchen293 (approved):
lgtm again, thanks for the first contribution!

nwithan8 (approved):

PR #403, opened 2023-08-09
fix: typescript definitions for lowestRate functions

### LINKED ISSUE #395: [Bug]: Type definition of `Shipment.lowestRate` function should not be static
### Software Version

6.6.1

### Language Version

v16.15.1

### Operating System

MacOS Ventura 13.3.1 (a)

### What happened?

The Typescript type definition of `Shipment.lowestRate` defines the function as being static: https://github.com/EasyPost/easypost-node/blob/master/types/Shipment/Shipment.d.ts#L285

In reality, the `Shipment.lowestRate` is not a static function: https://github.com/EasyPost/easypost-node/blob/2c07ac2c727dcffe23eef1543b39fe68a2a65afa/src/models/shipment.js#L44

### What was expected?

The expectation would be that the Typescript type definition defines the `Shipment.lowestRate` function as not being static.

### Sample Code

_No response_

### Relevant logs

_No response_

Justintime50:
# Description

It was reported that the Typescript definitions for `lowestRate` are still static when in reality they are the only functions that are no longer static. This PR fixes the definitions and adds the missing ones for Orders/Pickups.

Closes #395

# Testing

# Pull Request Type

### DIFF
--- a/types/Order/Order.d.ts
+++ b/types/Order/Order.d.ts
@@ -129,7 +129,7 @@ export declare class Order implements IOrder {
    *
    * @param carriers a list of carriers to filter rates for.
    * @param services a list of services to filter rates for.
-   * @returns {Promise<Rate>} The lowest {@link Rate}.
+   * @returns {Rate} The lowest {@link Rate}.
    */
   lowestRate(carriers?: string[], services?: string[]): IRate;
 }
--- a/types/Pickup/Pickup.d.ts
+++ b/types/Pickup/Pickup.d.ts
@@ -177,7 +177,7 @@ export declare class Pickup implements IPickup {
    *
    * @param carriers a list of carriers to filter rates for.
    * @param services a list of services to filter rates for.
-   * @returns {Promise<Rate>} The lowest {@link Rate}.
+   * @returns {Rate} The lowest {@link Rate}.
    */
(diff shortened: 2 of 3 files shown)


types/Order/Order.d.ts
```diff
@@ -123,4 +124,12 @@ export declare class Order implements IOrder {
...
    */
   static getRates(orderId: string): Promise<IRate[]>;
+
+  /**
+   *
+   * @param carriers a list of carriers to filter rates for.
+   * @param services a list of services to filter rates for.
+   * @returns {Promise<Rate>} The lowest {@link Rate}.
+   */
+  lowestRate(carriers?: string[], services?: string[]): IRate;
```
ralexmatthews, line 134: Issue:
Hmmm... it says it returns a promise of `Rate`, but the type says it just returns `IRate`. Should this be `Promise<IRate>` instead?

ralexmatthews (changes requested):
Niiiice. Just a small fix needed for these

types/Order/Order.d.ts
Justintime50: Ooo it's not a promise since we're just filtering a list. I'll fix the returns comment instead

ralexmatthews (approved):
Nice work dude 👏

nwithan8 (approved):

PR #422, opened 2023-11-16
Forward params on pagination

ralexmatthews:
# Description

There was an issue where sometimes when fetching the next page, we wouldn't have the original params sent along. So this stores the data used when fetching the next page in a new key on the EasyPostObject, `_params`, that is then added when fetching the next page.

# Testing

I ran the test suite and it worked, with some modifications to make sure we can ignore the _params when comparing input to output.

# Pull Request Type


test/helpers/utils.js
```diff
@@ -0,0 +1,27 @@
...
+          : Array.isArray(value)
+          ? [
+              key,
+              value.map((arrValue) =>
+                typeof arrValue === 'object' ? withoutParams(arrValue) : arrValue,
+              ),
+            ]
+          : typeof value === 'object'
+          ? [key, withoutParams(value)]
+          : [key, value],
```
jchen293, line 24: This function is very hard to read, can we split them into multiple conditional branches?

ralexmatthews: Fixed

nwithan8 (approved):
Paired with Alex on this, can confirm logic seems sound, approach actually makes it a bit more future-proof and resilient while also cleaning up bad automagical stuff.

PR #438, opened 2024-03-16
Fix incorrect types

bensontrent:
I made zip optional in IAddressCreateParameters since a zip code is not required in many countries.  Also added optional service and tax_identifiers to IShipmentCreateParameteres to enable one-call buys.

### DIFF
--- a/types/Address/Address.d.ts
+++ b/types/Address/Address.d.ts
@@ -36,7 +36,7 @@ export declare interface IAddress extends IObjectWithId<'Address'> {
   /**
    * ZIP or postal code the address is located in
    */
-  zip: string;
+  zip?: string;
 
   /**
    * ISO 3166 country code for the country the address is located in
@@ -99,7 +99,7 @@ export declare class Address implements IAddress {
   street2?: string | null;
   city?: string | null;
   state?: string | null;
-  zip: string;
+  zip?: string;
   country: string;
   residential?: boolean | null;
   carrier_facility?: string | null;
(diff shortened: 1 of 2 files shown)


Justintime50 (changes requested):
Thanks for submitting this, we'll need a couple changes to be able to roll this in:

1. Although zip codes are not required for some countries, our API does expect zip codes presently for addresses. You will want to work with our support team to determine what may need to go in the zip code field for countries that don't support it. As such, we cannot accept the nulling of the zip fields on the address object as it would not be a correct change to make.
2. With TaxIdentifier being a top-level object of our API, it would be better to place the type inside of its own folder structure like the other type objects are and follow the convention for defining it (with documenting the params as well like other objects).

If you'd rather not make these adjustments, let us know and we can continue the work you've started. Service and tax identifiers will be great additions that were previously missed.

bensontrent:
You can close this pull release, I'm just offering it as a suggestion. 

Both zip and country are optional in IAddressCreateParameters according to the actual way the API works:  the API will correctly return an error but the API does technically accept a blank value.    The reason a zip ought to be optional: consider a new user-facing address form, all of those values would initially be blank.  If my form object has a typescript type of IAddressCreateParameters, I'd have errors in my code since a zip code would initially be blank.  Also we, know a street1 and a name are required to create a shipment, but having those values as optional seem to be correct way to handle the creation of a form object while the form is in the process of being filled out.

The changes I submitted were necessary for my typescript to validate, so I'll just use my own fork of the project to facilitate the needs of my project.  

Thanks for looking at this!

PR #442, opened 2024-04-05
Import types with the type-only import

vassalloandrea:
# Description

Fix the TS error after upgrading to the version 5.

Error: 'ParametersToOmitOnCreate' is a type and must be imported using a type-only import when 'verbatimModuleSyntax' is enabled.

Error: 'IAddress' is a type and must be imported using a type-only import when 'verbatimModuleSyntax' is enabled.

image

# Pull Request Type


Justintime50:
I'm going to close this in favor of #488 which should still get you the desired results. Let me know if that's not the case with the patch version coming out shortly.

PR #444, opened 2024-04-05
[bug] Cannot determine payment method when funding/deleting

nwithan8:
# Description

Newer payment method objects have `pm_` prefixes for their IDs rather than `card_` or `bank_`, meaning we can no longer rely on the prefix to determine the payment method type (to determine the endpoint needed for funding and deleting a payment method).

This now falls back to the `object` key ("CreditCard" or "BankAccount") to determine the payment method type.

# Testing

- Unit tests for funding and deleting payment methods are mocked, so we can't test this E2E.
  - Add unit test to confirm switch-case working for type determination when prefix is `pm_`

# Pull Request Type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Next Release
+
+- Fix payment method funding and deletion failures due to undetermined payment method type
+
 ## v7.2.0 (2024-01-22)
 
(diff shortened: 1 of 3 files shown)


Justintime50 (approved):
thought: it's probably a good idea we test both implementations of ID and object name to ensure we don't regress - not just the new approach.

jchen293 (approved):

jchen293 (approved):

PR #455, opened 2024-07-09
feat: Account for new UPS endpoint in carrier account create/update functions

nwithan8:
# Description

- Routes `UpsAccount`, `UpsMailInnovationsAccount`, and `UpsSurepostAccount` create/update requests to the new `/ups_oauth_registrations` endpoint

# Testing

Adds new tests for each, updates previous custom workflow test to use FedEx instead of UPS

# Pull Request Type

### DIFF
--- a/src/services/carrier_account_service.js
+++ b/src/services/carrier_account_service.js
@@ -1,7 +1,6 @@
 import baseService from './base_service';
 import Constants from '../constants';
 import InvalidParameterError from '../errors/general/invalid_parameter_error';
-import InvalidRequestError from '../errors/api/invalid_request_error';
 
 const util = require('util');
 
@@ -27,7 +26,7 @@ export default (easypostClient) =>
       }
 
       const endpoint = this._selectCarrierAccountCreationEndpoint(carrierAccountType);
-      const wrappedParams = this._wrapCarrierAccountCreationParams(carrierAccountType, params);
+      const wrappedParams = this._wrapCarrierAccountParams(carrierAccountType, params);
 
       return this._create(endpoint, wrappedParams);
     }
--- a/src/services/carrier_account_service.js
+++ b/src/services/carrier_account_service.js
@@ -42,16 +42,10 @@ export default (easypostClient) =>
     static async update(id, params) {
       const carrierAccount = await this.retrieve(id);
 
-      if (!carrierAccount) {
-        throw new InvalidRequestError({
-          message: util.format(Constants.NO_OBJECT_FOUND, 'carrier account'),
-        });
-      }
-
       const carrierAccountType = carrierAccount.type;
 
       const endpoint = this._selectCarrierAccountUpdateEndpoint(carrierAccountType, id);
-      const wrappedParams = this._wrapCarrierAccountUpdateParams(carrierAccountType, params);
+      const wrappedParams = this._wrapCarrierAccountParams(carrierAccountType, params);
 
       try {
         const response = await easypostClient._patch(endpoint, wrappedParams);
@@ -97,20 +91,6 @@ export default (easypostClient) =>
     }
 
     /**
-     * Wraps the carrier account creation parameters in the correct format based on the type.
-     * @private
-     * @param {string} carrierAccountType - The type of carrier account to be created.
-     * @param {Object} params - The parameters for the carrier account to be created.
-     * @returns {Object} - The wrapped carrier account creation parameters.
-     */
-    static _wrapCarrierAccountCreationParams(carrierAccountType, params) {
-      if (Constants.UPS_OAUTH_CARRIER_TYPES.includes(carrierAccountType)) {
-        return { ups_oauth_registrations: params };
-      }
-      return { carrier_account: params };
-    }
-
-    /**
      * Returns the correct carrier_account endpoint when updating a record based on the type.
      * @private
      * @param {string} carrierAccountType - The type of carrier account to be updated.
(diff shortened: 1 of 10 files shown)


src/services/carrier_account_service.js
```diff
@@ -40,13 +40,21 @@ export default (easypostClient) =>
...
-      const wrappedParams = {
-        carrier_account: params,
-      };
+      const carrierAccount = await this.retrieve(id);
+
+      if (!carrierAccount) {
+        throw new InvalidRequestError({
+          message: util.format(Constants.NO_OBJECT_FOUND, 'carrier account'),
+        });
+      }
```
jchen293, line 49: We don't need this, our API will handle this when the carrier account doesn't exist

src/services/carrier_account_service.js
```diff
@@ -79,12 +87,57 @@ export default (easypostClient) =>
...
+     * @param {string} carrierAccountType - The type of carrier account to be updated.
+     * @param {Object} params - The parameters for the carrier account to be updated.
+     * @returns {Object} - The wrapped carrier account update parameters.
+     */
+    static _wrapCarrierAccountUpdateParams(carrierAccountType, params) {
+      if (Constants.UPS_OAUTH_CARRIER_TYPES.includes(carrierAccountType)) {
+        return { ups_oauth_registrations: params };
+      }
+      return { carrier_account: params };
+    }
```
jchen293, line 139: This function does the same thing as `_wrapCarrierAccountCreationParams` above, you can combine them into one function and rename it something like `_wrapCarrierAccountParams`

test/services/carrier_account.test.js
```diff
@@ -56,6 +57,23 @@ describe('CarrierAccount Service', function () {
...
+    const data = {
+      type: 'UpsAccount',
+      account_number: '123456789',
+    };
+
+    try {
+      const carrierAccount = await this.client.CarrierAccount.create(data);
+      await this.client.CarrierAccount.delete(carrierAccount.id);
+      // If we get here, the UPS account was created (and deleted) successfully
+      expect(true).to.equal(true);
```
jchen293, line 70: This assertion is meaningless. We should assert the `carrierAccount` attributes like what you have down below for update unit test

test/services/carrier_account.test.js
```diff
@@ -76,22 +94,50 @@ describe('CarrierAccount Service', function () {
...
+
+    // Update the carrier account
+    await this.client.CarrierAccount.update(carrierAccount.id, updateParams);
+
+    // Retrieve the updated carrier account
+    const updatedCarrierAccount = await this.client.CarrierAccount.retrieve(carrierAccount.id);
+
+    expect(updatedCarrierAccount).to.be.an.instanceOf(CarrierAccount);
+    expect(updatedCarrierAccount.id).to.match(/^ca_/);
+    // expect(updatedCarrierAccount.credentials["account_number"]).to.equal(testAccountNumber); // TODO: Re-enable assertion when API updates number properly
```
jchen293, line 140:
```suggestion
    // expect(updatedCarrierAccount.fields.credentials["account_number"]).to.equal(testAccountNumber); // TODO: Re-enable assertion when API updates number properly
```

There is another layer of key to access credentials, and I don't think we need this comment

src/services/carrier_account_service.js
nwithan8: Yes, but this cancels the function early if the carrier account doesn't exist.

src/services/carrier_account_service.js
nwithan8: Eh, I'd like to keep it explicit.

test/services/carrier_account.test.js
nwithan8: Resolved

test/services/carrier_account.test.js
nwithan8: `fields` is different than `credentials`. For now, both are empty because the API is not returning the proper data, hence the comment: 
image

src/services/carrier_account_service.js
jchen293: But these two methods have exactly the same logic and do the same thing.

src/services/carrier_account_service.js
Justintime50: Agreed, let's not add extra errors where they aren't needed. Our API handles these well. By introducing our own errors locally, we have to maintain them, we start to splinter the experience across libs when one works one way and another, another. It also means that the actual error messages users expect may differ since our lib message will be different from the API message on a 404 type error.

src/services/carrier_account_service.js
Justintime50: I feel this is a premature optimization. We can always break these out in the future if they differ. We should be opting to always keep code in it's most primitive and simple form and only expand and abstract when it makes sense which is not the case here.

test/services/carrier_account.test.js
```diff
@@ -56,6 +57,30 @@ describe('CarrierAccount Service', function () {
...
+  it('creates a UPS carrier account', async function () {
+    const accountNumber = '123456789';
+    const type = 'UpsAccount';
+
+    const data = {
+      type: type,
+      account_number: accountNumber,
+    };
+
+    try {
```
Justintime50, line 69: question: why do we have a try/catch here? This can obfuscate an actual problem in our test suite and should not be introduced. Let's just follow the other libs in how we test without getting too creative here. If we can't create and then immediately delete this account in our tests, we need to know and not bury it.

test/services/carrier_account.test.js
```diff
@@ -76,22 +101,50 @@ describe('CarrierAccount Service', function () {
...
+    // Create a temporary carrier account
+    const carrierAccount = await this.client.CarrierAccount.create(Fixture.basicCarrierAccount());
+
     const testDescription = 'My custom description';
+    const updateParams = {
+      description: testDescription,
+    };
 
-    const carrierAccount = await this.client.CarrierAccount.create(Fixture.basicCarrierAccount());
+    // Update the carrier account
```
Justintime50, line 112: style: let's not include comments that say what the next line does when it's easy to see that `create()` creates a record and so on. The only real comment that makes sense to keep is the one detailing why we need to delete a record in a test context when it's not the thing being tested. The other comments should be avoided.

test/services/carrier_account.test.js
Justintime50: let's remove this until it's a for-sure thing they're going to be fixing this. I'm not sure we can rely on a timetable from data.

Justintime50 (changes requested):

src/services/carrier_account_service.js
nwithan8: I see, misunderstanding on my part. If the API responds with a 4xx when attempting that, we're already set up to throw an exception, no need to double-catch. Got it, will remove.

Justintime50 (approved):

jchen293 (approved):
lgtm, thanks for addressing all the feedback!

PR #460, opened 2024-07-23
Adds support for claims API

ralexmatthews:
# Description

Creates the model, service, tests, and types for the claims API.

Also fixes a minor issue with the beta APIs where it added two slashes, making it like `//beta/whatever`

# Testing

Adds new tests based off the existing insurance tests

# Pull Request Type

### DIFF
--- a/test/services/claim.test.js
+++ b/test/services/claim.test.js
@@ -1,0 +1,103 @@
+/* eslint-disable func-names */
+import { expect } from 'chai';
+
+import EasyPostClient from '../../src/easypost';
+import EndOfPaginationError from '../../src/errors/general/end_of_pagination_error';
+import Claim from '../../src/models/claim';
+import Fixture from '../helpers/fixture';
+import * as setupPolly from '../helpers/setup_polly';
+
+/** @typedef {import("../../types/EasyPost").default} Client */
+
+const CLAIM_AMOUNT = 100;
+
+/** @param {Client} client */
+const buyTestShipment = async (client) => {
+  const shipment = await client.Shipment.create(Fixture.fullShipment());
+  const rate = shipment.lowestRate();
+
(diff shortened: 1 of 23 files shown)


test/services/claim.test.js
```diff
@@ -0,0 +1,103 @@
...
+      const claims = await client.Claim.all({ page_size: Fixture.pageSize() });
+      const nextPage = await client.Claim.getNextPage(claims, Fixture.pageSize());
+
+      const firstIdOfFirstPage = claims.claims[0].id;
+      const firstIdOfSecondPage = nextPage.claims[0].id;
+
+      expect(firstIdOfFirstPage).to.not.equal(firstIdOfSecondPage);
+    } catch (error) {
+      if (!(error instanceof EndOfPaginationError)) {
+        throw new Error('Test failed intentionally');
```
nwithan8, line 90: I believe this logic is backwards? Getting an EndOfPaginationError is fine and is the eventual expected path. Getting any other error is an error. Perhaps the wording "intentionally" is invalid here.

nwithan8 (approved):

nwithan8 (changes requested):
Please bump all endpoints to their v2 (GA) equivalent.

test/services/claim.test.js
ralexmatthews: Hmmm... I just copy-pasted this from insurance's tests. I'm not quite sure I understand? 

> Getting an EndOfPaginationError is fine and is the eventual expected path
 Yea, I think thats what this is counting on. If it's _not_ an `EndOfPaginationError`, _then_ throw an error.

I can probably refactor this to be more clear what we are doing with something like
```js
if (error instanceof EndOfPaginationError) {
  // all gucci
  return;
}

throw error
```

nwithan8 (approved):

Issue #456, opened 2024-07-11
[Bug]: Typescript types for exports don't have corresponding values


thyming:
### Software Version

7.3.1

### Language Version

20.13.1

### Operating System

OSX

### What happened?

1. `import { Tracker } from '@easypost/api'`;
2. `x instanceof Tracker`
3. `Uncaught TypeError: Right-hand side of 'instanceof' is not an object`

### What was expected?

`Tracker` and i'm sure other types should be exported from the root of the package, excluded from the root level export, or exported only as a type.

Justintime50:
Hey @thyming, thanks for submitting this. We are planning a Typescript rework here shortly that should hopefully address a lot of these issues. I don't yet have a timeframe but it's definitely on our roadmap.

thyming:
Sounds good, thanks!

Issue #446, opened 2024-04-15
[Bug]: TypeError: api_1.default is not a constructor in Nestjs


mustafakendiguzel:
### Software Version

7.3.0

### Language Version

21.4.0

### Operating System

Windows

### What happened?

1. Imported easypost-node library in nestjs
2. Attempt to create new EasyPostClient via 
```
import EasyPost from '@easypost/api';
export class ShippingMailingService {
  private readonly easyPostClient: EasyPost
  constructor() { 
    this.easyPostClient = new EasyPost('apiKey')
  }
}
```
3. Upon building, error is thrown that . TypeError: api_1.default is not a constructor
4. Tsconfig.json
```
{
  "compilerOptions": {
    "module": "commonjs",
    "declaration": false,
    "removeComments": true,
    "noLib": false,
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "allowSyntheticDefaultImports": true,
    "resolveJsonModule": true,
    "target": "es2017",
    "sourceMap": true,
    "allowJs": true,
    "outDir": "./dist",
    "baseUrl": "./",
    "lib": ["ES7", "ES2021", "DOM"],
    "incremental": true,
    "skipLibCheck": true,
    "strictNullChecks": false,
    "noImplicitAny": false,
    "strictBindCallApply": false,
    "forceConsistentCasingInFileNames": false,
    "noFallthroughCasesInSwitch": false,
  }
}
```

### What was expected?

To be able to build the app while using EasyPostClient

### Sample Code

```node
import EasyPost from '@easypost/api';

export class ShippingMailingService {
  private readonly easyPostClient: EasyPost
  constructor() { 
    this.easyPostClient = new EasyPost('apiKey')
  }
}

```

nwithan8:
Hello there, thanks for writing in!

I believe there's an issue with your provided code snippet, you should be trying to construct an `EasyPostClient` object, not an `EasyPost` object

```js
import EasyPostClient from '@easypost/api'; 

export class ShippingMailingService { 
  private readonly easyPostClient: EasyPostClient constructor() { this.easyPostClient = new EasyPostClient('apiKey') } }
```

mustafakendiguzel:
> Hello there, thanks for writing in!
> 
> I believe there's an issue with your provided code snippet, you should be trying to construct an `EasyPostClient` object, not an `EasyPost` object
> 
> ```js
> import EasyPostClient from '@easypost/api'; 
> 
[...]

PR #461, opened 2024-07-23
Removes Webpack/Babel in favor of SWC

ralexmatthews:
# Description

In its current state, the node library uses a webpack + babel build pipeline to bundle the whole library into one big CommonJS bundle. It uses maybe a dozen different webpack and babel settings, configs, plugins, etc to do this as well. This is probably not ideal for several reasons. 

- Since this should only be used in Node.js, not in the browser, this does not need bundled, so Webpack isn't needed altogether
- Babel, while popular, isn't super fast, and has lots of plugins needed to get the settings we want

Enter [SWC](https://swc.rs/), the Babel replacement written in Rust. It is a dependency of Turbopack, Vercel's new transpiler/bundler for Next.js, and seems to be the current future. 

As part of this move to SWC, I have also created both a CommonJS and ES6 build of the app to support the movement of the community away from CommonJS.

Getting rid of all the babel deps also removes these warnings when running `npm install`:
```
npm WARN deprecated @babel/plugin-proposal-export-namespace-from@7.18.9: This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-export-namespace-from instead.
npm WARN deprecated @babel/plugin-proposal-class-properties@7.18.6: This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-class-properties instead.
npm WARN deprecated @babel/plugin-proposal-unicode-property-regex@7.18.6: This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-unicode-property-regex instead.
npm WARN deprecated core-js@2.6.12: core-js@<3.23.3 is no longer maintained and not recommended for usage due to the number of issues. Because of the V8 engine whims, feature detection in old core-js versions could cause a slowdown up to 100x even if nothing is polyfilled. Some versions have web compatibility issues. Please, upgrade your dependencies to the actual version of core-js.
```

Overall, this greatly reduces the number of dependencies, config files, and improves support for CJS and ESM projects.

Also fixes https://github.com/EasyPost/easypost-node/issues/446. I was able to replicate the error by creating a new Nest app, and by changing the version to this, it fixes that error.

Also fixes https://github.com/EasyPost/easypost-node/issues/456 by exporting all errors and models from the root.

# Testing

All tests are still passing, thanks to swapping out babel/register to swc-node/register

# Pull Request Type


.github/workflows/ci.yml
```diff
@@ -14,14 +14,19 @@ jobs:
...
+          node-version: 20.x
       - run: make install
       - run: make install-styleguide
       - run: make build
       - run: EASYPOST_TEST_API_KEY=123 EASYPOST_PROD_API_KEY=123 make test
+      - name: Use Node.js ${{ matrix.node-version }} to test importing and using
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+      - run: make test-import-and-use
```
ralexmatthews, line 29: I don't think we actually care that you can build, test, and generally dev on the project in Node.js 12. This builds the app on a modern version, then uses the older version to run a small script that imports and attempts to use the lib.

.github/workflows/ci.yml
```diff
@@ -14,14 +14,19 @@ jobs:
         node-version: [12.x, 13.x, 14.x, 15.x, 16.x, 17.x, 18.x, 19.x, 20.x, 21.x, 22.x]
     steps:
       - uses: actions/checkout@v3
-      - name: Use Node.js ${{ matrix.node-version }}
+      - name: Use Node.js 20.x To Build
```
nwithan8, line 17: Why not Node 22 LTS?

nwithan8 (dismissed):

.github/workflows/ci.yml
ralexmatthews: Fair ¯\_(ツ)_/¯

nwithan8 (approved):

ralexmatthews:
Ah shoot. Messed up the merge.

nwithan8 (approved):

.github/workflows/ci.yml
```diff
@@ -14,14 +14,19 @@ jobs:
         node-version: [12.x, 13.x, 14.x, 15.x, 16.x, 17.x, 18.x, 19.x, 20.x, 21.x, 22.x]
     steps:
       - uses: actions/checkout@v3
-      - name: Use Node.js ${{ matrix.node-version }}
+      - name: Use Node.js 22.x To Build
```
Justintime50, line 17: suggestion: let's drop the version here as it'll inevitably get missed in a future update when the actual node version below gets bumped but the title gets forgotten.
```suggestion
      - name: Use Node.js To Build
```

package.json
```diff
@@ -11,14 +11,23 @@
...
   "license": "MIT",
   "engines": {
     "node": ">= 12.0"
   },
+  "exports": {
+    ".": {
+      "require": "./dist/cjs/index.js",
+      "import": "./dist/esm/index.js"
+    }
+  },
```
Justintime50, line 25: question: I assume this replaces the `index.js` file in the root? This accomplishes the same results though allowing users to require the entire library?

Justintime50 (dismissed):
praise: I'm ecstatic to see this! 😃 Solid work, babel and webpack have been the bane of this lib for years...

Curious, does this break anything? I looked into removing babel a couple years back and it got very angry at me, it's possible that over the last couple years and the various refactors we've done we've slowly ripped out anything that previously was a problem but did want to check if there were any implications. We did previously remove the ability for this to run in the browser which may have been the sole thing needed to unblock this.

package.json
ralexmatthews: Yea, exactly. Starting in Node v12, you can use the [conditional exports](https://nodejs.org/docs/latest-v22.x/api/packages.html#conditional-exports) to specify what they should get when they use your package. Basically for both the CJS and ESM builds, the entire src directory is built and built in these folders, so if they are using an `import EasyPostClient from "@easypost/api"` it will import the "import"/esm build, and for older projects that do a `const EasyPostClient = require("@easypost/api")` it will use the "require"/cjs built assets.

Justintime50 (approved):

PR #471, opened 2024-08-13
fix: webhook validation when float weight field is present (closes #467)

### LINKED ISSUE #467: [Bug]: HMAC encryption mismatch
### Software Version

7.5.0

### Language Version

Nodejs

### Operating System

Mac OS

### What happened?

The weight field in the tracker.updated EasyPost webhook is not properly formatted as JSON. When included, it appears as "weight": 136.0. However, when parsed as JSON, it is always reformatted to "weight": 136. This discrepancy leads to a mismatch of hashes when using the validateWebhook function.

As a result, about 50% of our webhooks are failing because some include a value for weight, while others have a value of null. When the weight field has a value of null, there is no issue with verifying the hash.

To work around this problem, we are using regex to insert the decimal point:

```
validateWebhook(
      Buffer.from(
        JSON.stringify(req.body).replace(/("weight":)(\d+)(?!\.\d)/g, "$1$2.0")
      ),
      headers["x-hmac-signature"],
      "12345"
    );
```
      
   

### What was expected?

Expected properly formatted JSON to be encrypted with the secret

### Sample Code

_No response_

### Relevant logs

_No response_

Justintime50:
# Description

When the `weight` field is present on a webhook (eg: tracking update), it gets converted but JS drops the decimal if `.0` which often is the case. This breaks webhook validation because HMAC signatures won't match if it's altered at all. This properly puts back the `.0` if necessary and missing to ensure that validation will continue to work. Tests validate that we properly compare the signatures

# Testing

# Pull Request Type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Next Release
+
+- Fixes webhook validation when the `weight` field contains a float by converting it back into a float after conversion from a string (closes #467)
+
 ## v7.5.1 (2024-08-09)
 
 - Adds missing properties to `Rate` model
--- a/src/utils/util.js
+++ b/src/utils/util.js
@@ -123,7 +123,7 @@ export default class Utils {
       // Fixes Javascript's float to string conversion. See https://github.com/EasyPost/easypost-node/issues/467
       const correctedEventBody = Buffer.from(eventBody)
         .toString('utf8')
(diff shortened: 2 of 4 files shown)


nwithan8 (dismissed):

jchen293 (dismissed):

Justintime50:
Fantastic, thanks for the updated info! We've incorporated that.

jchen293 (approved):

nwithan8 (approved):

PR #481, opened 2024-10-01
[chore] Prep for v7.5.3 release

nwithan8:
# Description

Bump version number, update changelog

# Testing

# Pull Request Type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # CHANGELOG
 
-## v7.6.0 (2024-10-01)
+## v7.5.3 (2024-10-02)
 
 - Removes the deprecated `create_list` tracker endpoint function as it is no longer available via API
 - Fix invalid TypeScript definition for Shipment `buy` function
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,8 +1,9 @@
 # CHANGELOG
 
-## Next Release
+## v7.6.0 (2024-10-01)
```
Justintime50, line 3: should be a patch version since it's only fixing something and not adding a new feature

nwithan8: Figured the removed function was worth a minor bump to hopefully cause caution for those updating.

Justintime50: That would improperly classify what's actually happening here though, those that have their libs pinned to a patch version won't get this update by doing a minor bump.

There are only a dozen customers using this feature behind a feature flag, all were contacted. It's also been removed today from the API so regardless of us removing it in the lib, it's no longer available to consume. At this point, it's a chore of removing dead code since the endpoint itself no longer exists.

Justintime50 (approved):

jchen293 (approved):

PR #485, opened 2025-01-02
Replaces Webpack/Babel/Mocha with Vite/Vitest

ralexmatthews:
# Description

Sorry, I was a dummy and accidentally closed #484 by deleting my own fork 🙃 so this is the exact same code, just on an easypost branch.

I decided to start with a much smaller problem to tackle, so this only removes Webpack/Babel/Mocha with Vite/Vitest. I chose Vite/Vitest because that it what [Easy-UI](https://github.com/EasyPost/easy-ui) uses for its builds and tests. 

There is a lot of files changed, but all changes are actually pretty minimal. All this does really is:

- Remove all webpack/babel/mocha dependencies and related files
- Installs Vite/Vitest and creates their config files
- Convert the tests to support Vitest
  - changes the way Polly is set up
  - remove all `this.` references in the tests
  - all cassettes work in place and none needed to change
- Makes some changes to github workflows
  - Adds a "test-node-compatibility" test that checks importing and using of the library
  - We don't want to just run the test suite because vite/vitest requires node v18+, but can generate output for arbitrarily low node versions

And thats effectively it. All commands still work as is. There are no changes to the source, and the output file is effectively the same. This does mean that it hasn't fixed any of the weird issues with using the project in Next.js and stuff. That can come at a later date since that will likely require breaking changes to fix.

Overall, this removes like half of the project's deps, and with it, the security vulnerabilities in exchange for a modern toolset.

# Testing

All unit tests are passing. In added another test/check specifically for checking compatibility with older node versions. It basically just imports the project and tries to create a test shipment. This makes it so we can test compatibility with old node versions while not requiring the toolchain to support old node versions.

```js
const EasyPostClient = require('../..');

const test = async () => {
  const client = new EasyPostClient('apiKey');

  if (!client.baseUrl.match(/^(http|https):\/\/api.easypost.com/gm)) {
    process.exit(1);
  }
};

test();
```

I also tested it with using actual API calls with a minimal project using node 12 and 22 and saw it working. I also checked node 10 and confirmed that did not work.

# Pull Request Type


nwithan8 (dismissed):

Justintime50:
@ralexmatthews if you can patch up the code scanning issue, I'll take a look

ralexmatthews:
@Justintime50 fixed 👍

test/node_compatibility/index.js
```diff
@@ -0,0 +1,11 @@
+const EasyPostClient = require('../..');
+
+const test = async () => {
+  const client = new EasyPostClient('apiKey');
+
+  if (!client.baseUrl.match(/^(http|https):\/\/api.easypost.com/gm)) {
```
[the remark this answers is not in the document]
Justintime50: @ralexmatthews hmmm, looks like another one popped up?

ralexmatthews:
Dang it lol

.github/workflows/ci.yml
```diff
@@ -8,16 +8,28 @@ on:
 
 jobs:
   build:
+    runs-on: ubuntu-latest
```
Justintime50, line 11: question: why are we breaking out the build and node-compatibility jobs into separate jobs? Seems that they are the same and therefore redundant?

package.json
```diff
@@ -43,25 +44,13 @@
...
   "devDependencies": {
-    "@babel/core": "^7.21.8",
-    "@babel/eslint-parser": "^7.21.8",
-    "@babel/plugin-proposal-class-properties": "^7.18.6",
-    "@babel/plugin-proposal-export-default-from": "^7.18.10",
-    "@babel/plugin-proposal-export-namespace-from": "^7.18.9",
-    "@babel/plugin-transform-react-constant-elements": "^7.21.3",
-    "@babel/plugin-transform-react-inline-elements": "^7.21.0",
-    "@babel/preset-env": "^7.21.5",
-    "@babel/register": "^7.21.0",
```
Justintime50, line 54: praise: I couldn't be more excited for this

test/helpers/setup_polly.js
```diff
@@ -1,9 +1,8 @@
-/* eslint-disable array-callback-return */
-/* eslint-disable no-param-reassign */
 import { resolve } from 'path';
 import NodeHttpAdapter from '@pollyjs/adapter-node-http';
-import { Polly, setupMocha as setupPolly } from '@pollyjs/core';
+import { Polly } from '@pollyjs/core';
 import FSPersister from '@pollyjs/persister-fs';
+// import { beforeAll, afterAll, beforeEach, afterEach } from 'vitest';
```
Justintime50, line 5: thought: let's remove commented, unused code

vite.config.js
```diff
@@ -0,0 +1,36 @@
+import { defineConfig } from 'vite';
+import path from 'path';
+
+const mode = process.env.NODE_ENV || 'development';
+const isDev = mode === 'development';
+
+export default defineConfig({
+  build: {
+    target: 'node12',
```
Justintime50, line 9: suggestion: I'd maybe leave a comment here detailing that we should target the oldest version of node we support so someone doesn't think it needs to be in inverse in the future and swap to say 22.

Justintime50 (commented):
Last time we attempted to overhaul the build system, we ran into issues once the final product was built, it couldn't be imported/referenced properly. Have we ensured that that issue is not a problem after these changes by running and E2E test by trying to import and use the final output of Vite?

.github/workflows/ci.yml
ralexmatthews: So they are not quite the same. The problem we are trying to solve here is that someone with node 12 installed cannot _build_ the project. Vite requires node 18+ to build. But we still want to support someone with node 12 installing and using the _built_ version of the app.

So we now have two tasks:
- `build` will test that the project builds and the test passes, but runs only on node 22 (or whatever is latest)
- `node_compatibility` will check that the built output can still be imported and used on older node versions.

ralexmatthews: I guess we could add the `node_compatibility` section after the build/test step of the `build` task 🤔 I could go either way

test/helpers/setup_polly.js
ralexmatthews: Good call.

.github/workflows/ci.yml
Justintime50: Ah I see, nice! I'd maybe just clarify this as comments above each job?

Justintime50 (approved):

ralexmatthews:
> Have we ensured that that issue is not a problem after these changes by running and E2E test by trying to import and use the final output of Vite?

Yea, I tested it with using actual API calls with a minimal project using node 12 and 22 and saw it working. I also checked node 10 and confirmed that did not work.

Also, thats what the node-compatibility CI check is supposed to check for. 

So as far as I can tell, it has full compatibility with the current setup

nwithan8 (approved):

PR #489, opened 2025-01-22
chore: prep v7.5.4 for release

### LINKED ISSUE #487: [Bug]: Enabling `verbatimModuleSyntax` breaks typechecking
### Software Version

7.5.2

### Language Version

Node 20.15.1, TS 5.5.3

### Operating System

macOS

### What happened?

I turned on `verbatimModuleSyntax` in my TSConfig, this surfaced an error from the `types` directory about not importing types using the `import type ...` syntax. This was specificly from the file `@easypost/api/types/Address/AddressCreateParameters.ts`.

The fix I got to work on my end was to patch the package, changing the erroring file to be a declaration file, i.e. 

`AddressCreateParameters.ts` -> `AddressCreateParameters.d.ts`. 

This matches the rest of the files under `types`, so it may have been a small oversight.

There seems to be an [open draft PR](https://github.com/EasyPost/easypost-node/pull/442) that fixes this, albeit differently than I am proposing (though, I believe my solution is more in line with the rest of the codebase). 

I have opened my own PR for this: 
- #488 

### What was expected?

No TypeScript error

Justintime50:
# Description

- Replaces Webpack with Vite for building the distributable library
- Fixes the filename for the `AddressCreateParameters` type declaration file (closes #487)
- Bumps dependencies

# Testing

# Pull Request Type

### DIFF
--- a/test/node_compatibility/index.js
+++ b/test/node_compatibility/index.js
@@ -1,4 +1,4 @@
-const EasyPostClient = require('../../dist/easypost');
+const EasyPostClient = require('../..');
 
 const test = async () => {
   const client = new EasyPostClient('apiKey');
--- a/test/node_compatibility/index.js
+++ b/test/node_compatibility/index.js
@@ -1,4 +1,4 @@
-const EasyPostClient = require('../..');
+const EasyPostClient = require('../../dist/easypost');
 
 const test = async () => {
   const client = new EasyPostClient('apiKey');
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## v7.5.4 (2025-01-22)
+
+- Replaces Webpack with Vite for building the distributable library
+- Fixes the filename for the `AddressCreateParameters` type declaration file (closes #487)
+- Bumps dependencies
+
 ## v7.5.3 (2024-10-02)
 
 - Removes the deprecated `create_list` tracker endpoint function as it is no longer available via API
--- a/index.js
+++ b/index.js
@@ -1,1 +1,1 @@
-module.exports = require('./dist/easypost.js').default;
+module.exports = require('./dist/easypost.js');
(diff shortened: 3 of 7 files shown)


ralexmatthews (dismissed):
Nice. LGTM 👍

ralexmatthews:
Ah just noticed the failing tests... checking that out...

test/node_compatibility/index.js
```diff
@@ -1,4 +1,4 @@
-const EasyPostClient = require('../..');
+const EasyPostClient = require('../../dist/easypost');
```
ralexmatthews, line 1: Hmmm... I don't think this is the way to fix this. It's explicitly importing from the root, because thats how it will be imported in those libraries. Doing it this way would be the equivalent of
```
const EasyPostClient = require("easypost/dist/easypost");
```
that would work, but is not ideal

Justintime50: Huh, when I hovered over it locally it kept telling me it was trying to import the TS file instead of the actual built lib. Is there an alternative I should look at to make it more explicit? Are you able to see the failed build errors?

ralexmatthews: Hmmm... yea thats what I think the `.d.ts` actually does. Thats where its going to get its types from. I think the `package.json` `main` field tells it to use the index file to get the actual code. Thats how its been working for the past while.

I can look into this if you'd rather. If this version needs to go out sooner, we can just omit the "fix the build errors" part of it for now while I'm looking at it

ralexmatthews: Ah nvm, figured it out.

Since we got rid of the mixed-name imports, it no longer needs the `.default` in the `index.js` file. Just removing that fixed it for me. 

I can push up a fix to this branch if you'd like

Justintime50: Sure! Thanks for checking on this. I assume that end users won't need to change anything since they're referencing the index file ultimately, we're just changing what's referenced inside of it?

ralexmatthews: That is correct. They are just importing the export from the index file, and don't care whats under the hood 👍

Justintime50: Fantastic, mind stamping this then? 😃

PR #492, opened 2025-02-05
Types fix

ralexmatthews:
# Description

This fixes an issue having to do with the type defs. The problem was since the app was changed to not have a `.default` export anymore, we couldn't use that in the type defs. 

This updates it to support the default export in both CJS and ESM syntax

CleanShot 2025-02-05 at 10 25 02@2x
CleanShot 2025-02-05 at 10 25 17@2x

# Testing

This is just a types change, so there should be no issue with the code itself.

I tested the code worked by creating a new node project and adding the project as a dep and checking the types were there.

# Pull Request Type


Justintime50 (approved):

matAlmeida:
Thank you @ralexmatthews. There is any ideia when this fix will be released?

PR #545, opened 2026-02-28
Enhance FieldError with code and suggestion properties

vCiKv:
# Description

Adds `IAddressVerificationFieldError` interface and replaces the `errors` key of `Verification` with it as it's the proper type. Adds missing `suggestion` to `IFieldError`

# Testing

 Just added optional types

# Pull Request Type

### DIFF
--- a/types/errors/FieldError.d.ts
+++ b/types/errors/FieldError.d.ts
@@ -3,9 +3,16 @@ export declare interface IFieldError {
    * Field of the request that the error describes
    */
   field: string;
-
   /**
    * Human readable description of the problem
    */
   message: string;
+  /**
+   * Error code of the request
+   */
+  code?: string;
+  /**
+   * possible error fix if available 
+   */
+  suggestion?: string;
 }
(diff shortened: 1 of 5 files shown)


types/errors/FieldError.d.ts
```diff
@@ -3,9 +3,16 @@ export declare interface IFieldError {
    * Field of the request that the error describes
    */
   field: string;
-
   /**
    * Human readable description of the problem
    */
   message: string;
+  /**
```
Justintime50, line 12: please include line breaks between each new entry

Justintime50 (changes requested):
Thanks for the PR! 

I think what I'd like to see here instead is the introduction of the `AddressVerificationFieldError` class (which has a code) since `FieldError` does not.

See https://docs.easypost.com/docs/errors and another related PR https://github.com/EasyPost/easypost-php/commit/e9ed0f6eb1330223a2e0760c7df68a4f8c916fab for more details. Ultimately it looks like we didn't properly  port those changes here when we did them so thanks for helping fill this gap!

types/errors/FieldError.d.ts
vCiKv: added the `AddressVerificationFieldError`  and reverted `FieldError` in new commit

Justintime50 (approved):
Thanks for the PR!

PR #548, opened 2026-06-03
Replace extractions/setup-just with EasyPost/ep-actions setup/just action

sroomberg-ep:
## What

Replaces the unverified third-party action `extractions/setup-just` with the internal composite action [`EasyPost/ep-actions/.github/actions/setup/just@main`](https://github.com/EasyPost/ep-actions/tree/main/.github/actions/setup/just).

## Why

`extractions/setup-just` is a third-party action from an unverified publisher and is **not** on the approved list in `EasyPost/ep-actions/ApprovedThirdPartyActions.yaml`. Using the internal composite action keeps CI on a vetted, EasyPost-owned action.

## Notes

- The ep-actions `setup/just` action installs Just on Linux runners; all affected jobs in this repo run on `ubuntu-latest`.
- The `just-version` input is supported by the ep-actions action where it was previously set.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
-      - uses: extractions/setup-just@v3
+      - uses: EasyPost/ep-actions/.github/actions/setup/just@main
       - uses: actions/setup-node@v6
         with:
           node-version: 24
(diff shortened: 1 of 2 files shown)


sroomberg-ep:
Backing out: `EasyPost/ep-actions` is a private repo and GitHub does not allow public repositories to consume actions from private repos (CI fails with "Unable to resolve action \`easypost/ep-actions\`, not found"). This repo will instead get an approved-list exception for `extractions/setup-just`. Closing in favor of that path.

PR #555, opened 2026-07-30
!feat: use fetch instead of superagent (closes #552)

### LINKED ISSUE #552: [Feat]: Migrate HTTP transport from superagent to fetch for edge-runtime compatibility
### Feature Request Is New

- [x] I have verified that the requested feature does not already exist or has not already been requested.

## Problem

The client can't run on edge runtimes (Cloudflare Workers, Vercel Edge, Deno Deploy) because:

- All requests go through superagent, which relies on node:http/node:https (Node) or XHR (browser) — neither exists in these runtimes; they only provide fetch.
- src/easypost.js builds the default User-Agent from os.platform()/os.release()/process.versions.node at module-eval time, so even importing the package throws.

## Proposal

Replace superagent with the standard fetch API (built into Node ≥18, all browsers, and every edge runtime).

The coupling is well-contained: services only call _request() and consume response.body/status/headers, so a fetch-based _request returning the same shape leaves services/, models/, and errors/ untouched. Sketch:

- URL/query via URL + URLSearchParams, Basic auth via an Authorization header, timeout via AbortSignal.timeout()
- Static User-Agent string (no os/process)
- Breaking: superagentMiddleware/requestMiddleware/useProxy expose superagent objects — replace with an injectable fetch option (covers proxying and mocking); requestHooks/responseHooks already use plain objects and survive as-is
- Drops the superagent dependency and shrinks the install

## Benefits

- Runs everywhere: Node ≥18, browsers, Cloudflare Workers, Vercel Edge, Deno, Bun — one transport, zero polyfills.
- Fewer dependencies: drops superagent and its transitive chain (formidable, hexoid, …), shrinking install size and supply-chain surface.
- Eliminates a known bug class: #439 (Next.js build failure via superagent → formidable → hexoid) was a symptom of this chain; fetch removes the class of problem rather than patching around it.
- No bundler workarounds: users stop needing webpack overrides to build in Next.js and similar frameworks.
- Future-proof: fetch is the platform standard; superagent-era conveniences (JSON parsing, query strings, timeouts) are each a few lines over fetch.

Happy to contribute a PR if there's interest.

Justintime50:
# Description

- Drops Node 16 support as `fetch` is only in Node 18+
- Breaking: HTTP transport migrated from `superagent` to fetch-compatible transport
  - `superagent` runtime dependency removed
  - Breaking: `superagentMiddleware` renamed to `httpMiddleware`
  - Breaking: `fetchClient` renamed to `httpClient`
  - Breaking: `makeApiCall` now accepts `delete` (the `del` alias was removed)
  - `requestMiddleware` compatibility preserved using a fetch-era compatibility request object
  - Default `User-Agent` retains structured runtime metadata fields (`Nodejs/`, `OS/`, `OSVersion/`, `OSArch/`) via runtime-safe detection

Closes #552 

# Testing

# Pull Request Type


UPGRADE_GUIDE.md
```diff
@@ -30,6 +31,42 @@ const boughtShipment = await client.Shipment.buy(shipment.id, shipment.lowestRat
...
+
+- Existing service APIs are unchanged.
+- `requestMiddleware` is still supported via a compatibility request object.
+
+What changed:
+
+- `superagent` is no longer a runtime dependency.
+- Node 18+ is now required (built-in `fetch`).
+- `superagentMiddleware` has been removed and replaced by `httpMiddleware`.
+- `fetchClient` has been renamed to `httpClient`.
```
copiousfreetime: Should this be `agent` has been renamed to `httpClient`?

src/easypost.js
```diff
@@ -60,14 +58,19 @@ export default class EasyPostClient {
     this.key = key;
     this.timeout = timeout || EasyPostClient.DEFAULT_TIMEOUT;
     this.baseUrl = baseUrl || EasyPostClient.DEFAULT_BASE_URL;
-    this.agent = superagent;
+    this.httpClient =
```
copiousfreetime, line 61: This is the rename that you're talking about in the Upgrade Guide right ?

test/services/address.test.js
```diff
@@ -179,11 +179,9 @@ describe('Address Service', function () {
     const addressData = Fixture.incorrectAddress();
 
     addressData.verify_carrier = 'UPS';
-    const address = await client.Address.createAndVerify(addressData);
-
-    expect(address).to.be.an.instanceOf(Address);
 
-    expect(address.verifications.delivery.errors[0].message).to.equal('Address not found');
-    expect(address.verifications.zip4.errors[0].message).to.equal('Address not found');
+    return client.Address.createAndVerify(addressData).catch((err) =>
```
copiousfreetime, line 183: This test changes what it tests -- its now checking that there is an exception on create and Verify instead of the address being a valid object, and the errors are in that object. Is this a behavior change in the api?

test/services/carrier_account.test.js
```diff
@@ -32,32 +32,6 @@ describe('CarrierAccount Service', function () {
     await client.CarrierAccount.delete(carrierAccount.id);
   });
 
-  it('creates a carrier account with a custom workflow', async function () {
```
copiousfreetime, line 35: Is there a reason for this test removal?

CHANGELOG.md
```diff
@@ -2,7 +2,15 @@
 
 ## v9.0.0 (Unreleased)
 
+- Breaking: Node 18+ is now required (built-in `fetch`)
 - Breaking: API resource responses are now plain JSON-compatible objects rather than model class instances
+- Breaking: HTTP transport migrated from `superagent` to fetch-compatible transport
+  - `superagent` runtime dependency removed
+  - Breaking: `superagentMiddleware` renamed to `httpMiddleware`
+  - Breaking: `fetchClient` renamed to `httpClient`
```
copiousfreetime: Should be `agent` renamed to `httpClient` ?

copiousfreetime (approved):
Just a few non-blocking clarifications.

test/services/carrier_account.test.js
Justintime50: Yes, FedEx no longer uses this path and returns errors as a result. An understood change.

test/services/address.test.js
Justintime50: Yes, something about our verification engine changed that this address now cannot be verified (known)

PR #557, opened 2026-08-06
TSM-01: Add TS scaffolding, project-local ESLint, and CI alignment

Justintime50:
## Summary
Establishes topology item 1 migration scaffolding for mixed JS/TS validation while keeping runtime behavior unchanged.

## Changes in this PR
- adds project-local `.eslintrc` with TypeScript parser override for `*.ts` files.
- updates `.github/workflows/ci.yml` install commands to align with project-local lint configuration.
- adds `tsconfig.base.json` and `tsconfig.build.json` for source migration scaffolding.
- updates `tsconfig.json` to extend the shared base config.
- updates `package.json` TypeScript script wiring for declaration/source checks.
- removes `.eslintrc` ignore from `.gitignore` so config is tracked.

## Testing
- `npm run typescript`
- `npx eslint test/types/permissive_types.ts`

## Stack Context
Bottom PR in stack #560.

PR #558, opened 2026-08-06
TSM-02: Add CJS/ESM export compatibility test

Justintime50:
## Summary
Implements topology item 2 by adding explicit package export compatibility coverage for both CJS and ESM consumption paths.

## Changes in this PR
- adds `test/services/module_exports_compat.test.js`.
- validates CommonJS `require('../..')` returns constructable client.
- validates ESM `import('../..')` default export returns constructable client.

## Testing
- `npx vitest run test/services/module_exports_compat.test.js`

## Stack Context
Depends on PR #557.

### DIFF
--- a/test/services/module_exports_compat.test.js
+++ b/test/services/module_exports_compat.test.js
@@ -1,0 +1,25 @@
+import { expect } from 'chai';
+import { createRequire } from 'module';
+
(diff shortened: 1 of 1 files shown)


ralexmatthews (approved):
🔥

samw-easypost (approved):

PR #561, opened 2026-08-06
TSM-04: convert core entry and shared infra to TypeScript

Justintime50:
## Summary
- rename core entry and shared modules to TypeScript: easypost, constants, util, internal_util
- update Vite library entry to src/easypost.ts and resolve extensions for .ts + .js
- keep runtime behavior unchanged while migration proceeds with temporary ts-nocheck on converted files
- allow permissive incremental migration in source typecheck by setting noImplicitAny=false in tsconfig.build.json

## Validation
- npm run typescript
- npm run build
- npx eslint src/easypost.ts src/constants.ts src/utils/util.ts src/utils/internal_util.ts
- npx vitest run test/services/module_exports_compat.test.js

### DIFF
--- a/src/easypost.ts
+++ b/src/easypost.ts
@@ -533,7 +533,11 @@ export default class EasyPostClient {
    * @param {Object} [headers] - Additional headers to send with the request.
    * @returns {*} The response from the HTTP request.
    */
-  _get(path: string, params: Record<string, unknown> = {}, headers: RequestHeaders = {}): Promise<any> {
+  _get(
+    path: string,
+    params: Record<string, unknown> = {},
+    headers: RequestHeaders = {},
+  ): Promise<any> {
     return this._request(path, EasyPostClient.METHODS.GET, params, headers);
   }
 
@@ -544,7 +548,11 @@ export default class EasyPostClient {
    * @param {Object} [headers] - Additional headers to send with the request.
    * @returns {*} The response from the HTTP request.
    */
-  _post(path: string, params: Record<string, unknown> = {}, headers: RequestHeaders = {}): Promise<any> {
+  _post(
+    path: string,
+    params: Record<string, unknown> = {},
+    headers: RequestHeaders = {},
+  ): Promise<any> {
     return this._request(path, EasyPostClient.METHODS.POST, params, headers);
   }
 
@@ -555,7 +563,11 @@ export default class EasyPostClient {
    * @param {Object} [headers] - Additional headers to send with the request.
    * @returns {*} The response from the HTTP request.
    */
-  _put(path: string, params: Record<string, unknown> = {}, headers: RequestHeaders = {}): Promise<any> {
+  _put(
+    path: string,
+    params: Record<string, unknown> = {},
+    headers: RequestHeaders = {},
+  ): Promise<any> {
     return this._request(path, EasyPostClient.METHODS.PUT, params, headers);
   }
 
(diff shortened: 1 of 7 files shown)


src/easypost.ts
```diff
@@ -39,14 +38,61 @@ import UserService from './services/user_service';
...
+type RequestHeaders = Record<string, string>;
+
+type ClientOptions = {
+  apiKey?: string;
+  useProxy?: boolean;
+  timeout?: number;
+  baseUrl?: string;
+  httpMiddleware?: any;
+  requestMiddleware?: any;
+  httpClient?: any;
```
ralexmatthews: Question:
Hmmm... are these just stubs for later, or are these actually anything? For example, `httpClient` could now be initialized with `2` and it would be valid according to TS. 

Ideally we would type these with what we expect. For example, `httpClient: typeof fetch` or similar.

tsconfig.build.json
```diff
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "allowJs": true,
     "checkJs": false,
+    "noImplicitAny": false,
```
ralexmatthews, line 6: Suggestion:
I would strongly argue to keep this enabled. This can very easily allow type errors to get swept under the rug. If you create a new function, this no longer requires that function's parameters to be typed. 
```typescript
const getFooBar = (arg) => arg.foo.bar;
// ...
getFooBar("abc")
```
now compiles fine, even though this will be a runtime error.

ralexmatthews (changes requested):
Nice, mostly looks good. I would just strongly suggest keeping `noImplicitAny` disabled here. 

Would also suggest stronger types on the client options, but that one is nonblocking.

sroomberg-ep (approved):

src/easypost.ts
Justintime50: This is more tightly typed now, good call. The first few PRs were too loose and I got more explicit as I went. Came back to patch this up.

tsconfig.build.json
Justintime50: We later remove this and tighten `any` some in #576, good call

src/easypost.ts
```diff
@@ -39,14 +38,82 @@ import UserService from './services/user_service';
...
+
+type ClientOptions = {
+  apiKey?: string;
+  useProxy?: boolean;
+  timeout?: number;
+  baseUrl?: string;
+  httpMiddleware?: HttpMiddleware;
+  requestMiddleware?: RequestMiddleware;
+  httpClient?: HttpClient;
+};
```
ralexmatthews, line 75: Siiiiick 🔥

src/easypost.ts
```diff
@@ -309,7 +386,12 @@ export default class EasyPostClient {
    * @returns {*} The response from the HTTP request.
    * @throws {ApiError} If the request fails.
    */
-  async _request(path = '', method = EasyPostClient.METHODS.GET, params = {}, headers = {}) {
+  async _request(
+    path = '',
+    method: string = EasyPostClient.METHODS.GET,
+    params: Record<string, unknown> = {},
+    headers: RequestHeaders = {},
+  ): Promise<any> {
```
ralexmatthews, line 394: question:
Do we wanna make this return `unknown` to be consistent with the others above? Or is that part of the "strengthening the any's" you referenced?

ralexmatthews (approved):
Looks good. Like the better clientoptions type. Looking forward to disabling that `noImplicitAnys` option though

src/easypost.ts
Justintime50: You may want to cross reference any questions with the end state, lots of changes kept happening up the stack too (unfortunate I know, but it's possible that these got tightened/corrected in a later PR)

PR #562, opened 2026-08-06
TSM-05: convert base service hydration layer to TypeScript

Justintime50:
## Summary
- rename base hydration files to TypeScript: base_service and easypost_object
- preserve dynamic model hydration and Symbol.hasInstance behavior unchanged
- keep migration permissive with temporary ts-nocheck on converted files

## Validation
- npm run typescript
- npm run build
- npx eslint src/services/base_service.ts src/models/easypost_object.ts
- npx vitest run test/services/module_exports_compat.test.js

## Notes
- targeted service tests that construct the client require EASYPOST_TEST_API_KEY in local env


src/models/easypost_object.ts
```diff
@@ -0,0 +1,32 @@
...
+  static _params: Record<string, unknown>;
+
+  static [Symbol.hasInstance](instance: unknown): boolean {
+    if (instance == null || typeof instance !== 'object') {
+      return false;
+    }
+
+    const modelConstructor = (instance as Record<symbol, unknown>)[
+      Symbol.for('easypost.modelConstructor')
+    ];
```
ralexmatthews: suggestion:
You can use a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) function here, so that way you don't need to keep casting. 
```typescript
const isSymbolRecord = (value: unknown) value is Record<symbol, unknown> =>
  instance != null && typeof instance === 'object';
// ...
static [Symbol.hasInstance](instance: unknown): boolean {
  if (!isSymbolRecord(instance)) {
    return false;
  }

  const modelConstructor = instance[Symbol.for('easypost.modelConstructor')];
```

ralexmatthews (approved):
Nice. Just one nit/suggestion

sroomberg-ep (approved):

src/models/easypost_object.ts
Justintime50: Good call, fixed

PR #563, opened 2026-08-06
TSM-06: convert Group A services to TypeScript

Justintime50:
## Summary
Converts Group A service modules to TypeScript and applies minimal typed compatibility fixes needed for the current base-service conversion, with runtime behavior preserved.

## Changes in this PR
- renames Group A service files to .ts:
  - src/services/address_service.ts
  - src/services/parcel_service.ts
  - src/services/customs_info_service.ts
  - src/services/customs_item_service.ts
  - src/services/shipment_service.ts
- adds minimal type-safe wrapper map annotations where property writes are dynamic
- updates conversion helper callsites to pass params consistently with current base service signatures
- introduces no new @ts-nocheck

## Testing
- npm run typescript
- npm run build
- npx eslint src/services/address_service.ts src/services/parcel_service.ts src/services/customs_info_service.ts src/services/customs_item_service.ts src/services/shipment_service.ts
- npx vitest run test/services/address.test.js test/services/parcel.test.js test/services/customs_info.test.js test/services/customs_item.test.js test/services/shipment.test.js test/services/module_exports_compat.test.js

## Notes
- targeted service suites above are blocked locally without EASYPOST_TEST_API_KEY; module export compatibility test is included in the command set.


src/services/address_service.ts
```diff
@@ -12,10 +33,10 @@ export default (easypostClient) =>
      * @param {Object} params - Parameters for the address to be created.
      * @returns {Address} - The created address.
      */
-    static async create(params) {
+    static async create(params: AddressCreateParameters): Promise<unknown> {
       const url = 'addresses';
 
-      const wrappedParams = {};
+      const wrappedParams: Record<string, unknown> = {};
```
ralexmatthews, line 41: nit:
Looks like this is a known type, and could probably strengthen it just a little bit.
```typescript
type MetaAddressKeys = "verify" | "verify_strict" | "verify_carrier";
type AddressCreateInternalParameters = 
  { address?: Omit<AddressCreateParameters, MetaAddressKeys> } & 
  Pick<AddressCreateParameters, MetaAddressKeys>;
```

src/services/address_service.ts
```diff
@@ -70,7 +91,7 @@ export default (easypostClient) =>
      * @param {Object} [params] - Parameters to filter the list of addresses.
      * @returns {Object} - An object containing a list of {@link Address addresses} and pagination information.
      */
-    static async all(params = {}) {
+    static async all(params: Record<string, unknown> = {}): Promise<unknown> {
```
ralexmatthews: nit:
Do we really not know what these return? It feels like these should be known.

ralexmatthews (commented):
Hmmm... this looks like its more of a "just adding types for the sake of adding types" and less like implementing typescript checking. Lots of unknowns, anys, and general type omissions. I see lots of places where we introduce more any's, and while we technically have "explicit parameter and return types", I don't think most of them are actually helpful. So many are just 
```typescript
(params: Record<string, unknown>): Promise<unknown>
```
Image

I might not be getting the point of this, and they are typed in the `./types/` directory. I don't know if we want to implement those in the source directly instead, or treat those as the source of truth or what, but it feels divergent.

samw-easypost (approved):

sroomberg-ep (approved):

src/services/address_service.ts
Justintime50: Yes, these hadn't made it into the PR before your review but should be there now.

src/services/address_service.ts
Justintime50: There's now create params for each function, most likely wasn't there yet when you began reviews.

ralexmatthews (approved):
Niiice. Still a handful of anys thrown around, but hoping those get fixed later

PR #564, opened 2026-08-07
TSM-07: convert Group B services to TypeScript

Justintime50:
## Summary
- Convert Group B services to TypeScript: Batch, Order, Pickup, Rate, SmartRate, ScanForm, and Refund.
- Keep runtime behavior unchanged while moving service modules to `.ts`.

## Validation
- npm run typescript
- npm run build
- Targeted Group B service tests


sroomberg-ep (approved):

samw-easypost (approved):

src/services/batch_service.ts
```diff
@@ -1,7 +1,13 @@
 import baseService from './base_service';
+import Batch from '../models/batch';
 
 export const DEFAULT_LABEL_FORMAT = 'pdf';
 
+type BatchCreateParameters = Record<string, unknown> & {
+  shipments?: Array<string | Record<string, unknown>> | null;
+};
```
ralexmatthews, line 8: thought:
These new types are still not as thorough as the old ones. For example, for the create batch params, here are the existing types in https://github.com/EasyPost/easypost-node/blob/master/types/Batch/BatchCreateParameters.d.ts:

```ts
export declare interface IBatchCreateParameters {
  shipments?: Array<Shipment | string>;
}
```

Where shipment is a fully fleshed type listing the fields and their types. 

I'm wondering if the plan is to eventually go through and use these existing types or if they should be reimplemented here? We are already importing the batch type. I wonder if we should import the create param here as well?

src/services/pickup_service.ts
```diff
@@ -1,4 +1,22 @@
 import baseService from './base_service';
+import Pickup from '../models/pickup';
+
+type PickupCreateParameters = Record<string, unknown> & {
+  address?: Record<string, unknown> | string | null;
```
ralexmatthews, line 5: thought:
Same thing here. This `Record<string, unknown>` is explicitly wider than what we had. Not sure if thats intentional. But then the sub-record fields are loosey goosey "just pass whatever", where the original types had
```ts
interface BasePickupCreateParameters {
  address: Address | string;
  // ...
```

ralexmatthews (approved):
Nice. Non blocking, but just want clairification on the intention of these PRs. Are we explicitly wanting all the input types to be much more loosey-goosey than before, or are we trying to like recreate/reuse the well defined and structured types from before?

PR #567, opened 2026-08-07
TSM-10: convert Group E services to TypeScript

Justintime50:
## Summary
- Convert Group E services to TypeScript: ApiKey, CustomerPortal, Embeddable, EndShipper, FedExRegistration, Luma, ReferralCustomer, and User.
- Preserve existing runtime behavior and public service interfaces.

## Validation
- npm run typescript
- npm run build
- Targeted Group E service tests


samw-easypost (approved):

src/services/api_key_service.ts
```diff
@@ -17,20 +25,20 @@ export default (easypostClient) =>
...
         if (user.id == id) {
-          return user.keys;
+          return user.keys ?? [];
         }
 
-        user.children.forEach((child) => {
+        user.children?.forEach((child) => {
           if (child.id == id) {
-            return child.keys;
+            return child.keys ?? [];
```
ralexmatthews: Issue:
This is unrelated to your changes here, but I don't think this would actually work the way it looks like is intended here. This `return child.keys ?? [];` only returns in the `forEach` arrow fn, not for the outside scope. If you wanted to return those child keys from the `retrieveApiKeysForUser` outside fn, you would need to either do like a `user.children?.find` and return that, or change this to like a `for (child in user.children ?? []) {` loop.

src/services/referral_customer_service.ts
```diff
@@ -3,16 +3,45 @@ import util from 'util';
...
+  recharge_amount?: string | null;
+  secondary_recharge_amount?: string | null;
+  recharge_threshold?: string | null;
+  children?: Record<string, unknown>[] | null;
+  api_keys?: Record<string, unknown>[] | null;
+};
+type MandateData = Record<string, unknown>;
+type ReferralCustomerListResponse = { referral_customers: User[]; has_more: boolean };
+/* eslint-disable no-unused-vars */
+type ReferralScopedClient = {
```
ralexmatthews: question:
Is this temporary, or is there a reason we are keeping this arround?

src/services/referral_customer_service.ts
```diff
@@ -24,12 +53,13 @@ function _getReferralClient(client, referralApiKey) {
...
  */
-async function _getEasyPostStripeKey(easypostClient) {
+async function _getEasyPostStripeKey(easypostClient: EasyPostHttpClient): Promise<string> {
   const url = 'partners/stripe_public_key';
 
   const response = await easypostClient._get(url);
 
-  return response.body.public_key;
+  const body = response.body as Record<string, unknown>;
+  return body.public_key as string;
```
ralexmatthews: thought:
If we are asserting the type anyway, we could shorten this to
```ts
const body = response.body as { public_key: string };
return body.public_key;
```

src/services/referral_customer_service.ts
```diff
@@ -83,12 +122,17 @@ async function _sendCardDetailsToStripe(stripeKey, number, expirationMonth, expi
  * @param {string} priority - Whether to add the card as the 'primary' or 'secondary' card.
  * @returns {Object} - Response body (EasyPost payment method object).
  */
-async function _sendCardDetailsToEasyPost(client, referralApiKey, stripeCreditCardToken, priority) {
+async function _sendCardDetailsToEasyPost(
+  client: EasyPostClient,
+  referralApiKey: string,
+  stripeCreditCardToken: string,
+  priority: string,
+): Promise<Record<string, unknown>> {
```
ralexmatthews: nit:
Unspecific return type here.

ralexmatthews (approved):
Few comments, nothing blocking

PR #568, opened 2026-08-10
TSM-11: convert Group A models to TypeScript

Justintime50:
## Summary
- Convert Group A models to TypeScript (Address, Parcel, CustomsInfo, CustomsItem, PostageLabel, Shipment).
- Add typed model field declarations using a permissive optional+nullable policy for migration safety.
- Include audit allowlist update needed to keep CI scan green.

## Validation
- npm run typescript
- npm run build
- Targeted Group A model/service tests


src/models/address.ts
```diff
@@ -0,0 +1,24 @@
+import EasyPostObject from './easypost_object';
+
+/**
+ * An {@link https://docs.easypost.com/docs/addresses Address} represents people, places, and organizations in a number of contexts.
+ * @public
+ * @extends EasyPostObject
+ */
+export default class Address extends EasyPostObject {
+  declare street1?: string | null;
```
sroomberg-ep, line 9: If `?` denotes an optional value why do you need to type these as `string | null` instead of just `string`?

sroomberg-ep: never mind I forgot about `undefined`

sroomberg-ep (approved):

src/models/shipment.ts
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * A {@link https://docs.easypost.com/docs/shipments Shipment} represents a physical {@link Parcel}, the origin and destination {@link Address Addresses}, and any associated {@link CustomsInfo}.
+ * @public
+ * @extends EasyPostObject
+ */
+export default class Shipment extends EasyPostObject {
+  declare batch_id?: string | null;
+  declare batch_message?: string | null;
+  declare batch_status?: string | null;
+  declare buyer_address?: Record<string, unknown> | null;
```
ralexmatthews, line 13: note:
got more of the `Record<string, unknown>` submodels here. Not sure if we want to use the nicer types later.

ralexmatthews (approved):

PR #569, opened 2026-08-10
TSM-12: convert Group B models to TypeScript

Justintime50:
## Summary
- Convert Group B models to TypeScript (Batch, Order, Pickup, PickupRate, Rate, Refund, ScanForm).
- Add permissive model field declarations (`?: ... | null`) to preserve compatibility during migration.

## Validation
- npm run typescript
- npm run build
- Targeted Group B model/service tests


sroomberg-ep (approved):

src/models/order.ts
```diff
@@ -0,0 +1,38 @@
...
+ * An {@link https://docs.easypost.com/docs/orders Order} represents a collection of packages, intended only for multi-parcel shipments.
+ * @public
+ * @extends EasyPostObject
+ */
+export default class Order extends EasyPostObject {
+  declare buyer_address?: Record<string, unknown> | null;
+  declare from_address?: Record<string, unknown> | null;
+  declare is_return?: boolean | null;
+  declare messages?: Record<string, unknown>[] | null;
+  declare rates?: Parameters<typeof Constants.Utils.getLowestRate>[0] | null;
```
ralexmatthews: Suggestion:
I wonder if we couldn't type this as something like `Rate` instead of trying to grab the param?

src/models/order.ts
```diff
@@ -0,0 +1,38 @@
...
+   * @param {string[]} [carriers] - List of allowed carriers to filter by
+   * @param {string[]} [services] - List of allowed services to filter by
+   * @returns {Rate} - The lowest rate
+   * @throws {FilteringError} - If no applicable rates are found
+   */
+  lowestRate(
+    carriers?: string[],
+    services?: string[],
+  ): ReturnType<typeof Constants.Utils.getLowestRate> {
+    const rates = ((this as Order & { rates?: unknown[] }).rates || []) as Parameters<
```
ralexmatthews: question:
Is this coercion needed? I would guess TS should pick this up

ralexmatthews (approved):

PR #576, opened 2026-08-24
TSM-92: enable noImplicitAny and harden typing

Justintime50:
## Summary
- enable noImplicitAny in build, declarations, and service-test tsconfigs
- resolve strictness fallout across source and service tests
- replace missing external type imports with local structural aliases where needed

## Validation
- npm run lint
- npm run formatCheck
- npm run typescript

### DIFF
--- a/src/services/base_service.ts
+++ b/src/services/base_service.ts
@@ -99,7 +99,7 @@ const RESOURCES = {
   Webhook,
 };
 
-export default (easypostClient) =>
+export default (easypostClient: any) =>
   /**
    * The base class for all EasyPost client library services.
    * @param {EasyPostClient} easypostClient The {@link EasyPostClient} instance to use for API calls.
@@ -112,13 +112,13 @@ export default (easypostClient) =>
      * @param {*} response The value to serialize.
      * @returns {*} A plain object/array/scalar.
      */
-    static _toPlainEasyPostObject(response) {
+    static _toPlainEasyPostObject(response: any): any {
       if (Array.isArray(response)) {
         return response.map((value) => this._toPlainEasyPostObject(value));
       }
 
       if (typeof response === 'object' && response !== null) {
-        const plainObject = {};
+        const plainObject: Record<string, any> = {};
         const prototype = Object.getPrototypeOf(response);
 
         if (prototype && prototype !== Object.prototype) {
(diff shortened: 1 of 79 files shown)


src/services/beta_referral_customer_service.ts
```diff
@@ -86,7 +90,9 @@ export default (easypostClient) =>
     static async createBankAccountClientSecret(
       returnUrl: string | null,
     ): Promise<BetaClientSecretResponse> {
-      const params = returnUrl ? { return_url: returnUrl } : null;
+      const params = returnUrl
+        ? { return_url: returnUrl }
+        : (null as unknown as Record<string, unknown>);
```
ralexmatthews: Issue:
Huh. Thats not right...

src/services/base_service.ts
```diff
@@ -173,23 +175,26 @@ export default (easypostClient) =>
       }
 
       if (typeof response === 'object' && response !== null) {
-        let classObject;
-        if (RESOURCES[response.object] !== undefined) {
-          classObject = new RESOURCES[response.object]();
+        const responseRecord = response as Record<string, any>;
```
ralexmatthews: suggestion:
We could use another type guard here to prevent the assertion

ralexmatthews (commented):
Hmmm... so this turns off `noImplicitAny`, but then it types a bunch of them as `any` anyway so we are not getting a lot of a wins of TS. 

In some of these place where stuff could be anything, like in the response checkers or errors, I think we should prefer `unknown`. `any` uses honor-system checking, "Yes TS, I promise I checked that this should work". `unknown` has TS check that it will work as expected. Its more verbose in places but a lot more type safe. 

You could also make the errors default to `unknown` with the [tsconfig setting](https://www.typescriptlang.org/tsconfig/#useUnknownInCatchVariables).

src/services/beta_referral_customer_service.ts
```diff
@@ -74,7 +75,10 @@ export default (easypostClient) =>
     static async createCreditCardClientSecret(): Promise<BetaClientSecretResponse> {
       const url = 'beta/setup_intents';
 
-      const response = await easypostClient._post(url, null);
+      // Preserve legacy null payload behavior for cassette matching in tests.
+      const emptyPayload = null as unknown as Record<string, unknown>;
+
+      const response = await easypostClient._post(url, emptyPayload);
```
ralexmatthews: suggestion:
Instead of saying null is a record, could we instead have the post request accept null as a body?

ralexmatthews (approved):
Niiiiice alot better 👏

sroomberg-ep (approved):

